PR #10, opened 2024-08-10
chore: add mysql tool

drpebcak:


clis/mysql/tool.gpt
```diff
@@ -0,0 +1,24 @@
+name: mysql
+type: context
+description: Provides the ability to interact with mysql databases.
+share tools: sys.exec
+share credential: github.com/gptscript-ai/credential as cli.mysql.password with MYSQL_PWD as env and password as field and "Please enter your mysql user's password:" as message
```
drpebcak, line 5: does this do what i think it does? im hoping this exposes the tool in such a way that it requires it for the consumer of this tool, but gives it a unique name on their system so that they could use this tool multiple times.

also, do we prefer writing a custom credential tool if we need multiple fields, or just having multiple credentials saved?

drpebcak: As a side note, it's a little annoying to test these when we have such different behavior when things are local vs from git - locally, 'share credential' seems to do nothing.

g-linville: Yeah, `Share Credential` works the same way as `Share Tools`, `Share Context`, etc.
The tool that has `tools: github.com/gptscript-ai/tools/clis/mysql` will also have this credential tool run. The "unique name on their system" is just the credential alias that you specified here (`cli.mysql.password`).

If you need multiple fields, I would recommend just using the generic credential tool multiple times. It's more convenient.

I'm surprised if `share credential` truly does nothing locally. Maybe there were some mismatching file paths that were an issue? Idk. Were you initially trying to use a custom credential tool you wrote locally rather than the generic one?

clis/mysql/tool.gpt
```diff
@@ -0,0 +1,24 @@
...
+
+#!/bin/bash
+
+if ! command -v mysql; then
+  echo "The user does not have the mysql cli installed or it is not available on the PATH."
+else
+  echo "The user has the mysql cli available. Use it to work with mysql."
+  echo "The following is the help text for the mysql cli. Always execute mysql statements non-interactively from the CLI using one of two options:"
```
g-linville, line 13: nit: this gets repeated down farther, so it doesn't need to be here
```suggestion
  echo "Always execute mysql statements non-interactively from the CLI using one of two options:"
```

g-linville (approved):

clis/mysql/tool.gpt
drpebcak: With this test script:
```
tools: github.com/drpebcak/tools/clis/mysql@mysql
chat: true
```
I would expect to get prompted to set up the credential when i call it, but it does not ask me to do anything. If i walk it through connecting to a db it just makes me type credentials into the chat session but those arent saved anywhere.

drpebcak: for clarity - it only asks for credentials in the context of needing them to connect to the db, not because its running the credential tool.

clis/mysql/tool.gpt
```diff
@@ -0,0 +1,24 @@
...
+share credential: github.com/gptscript-ai/credential as cli.mysql.password with MYSQL_PWD as env and password as field and "Please enter your mysql user's password:" as message
+
+#!/bin/bash
+
+if ! command -v mysql; then
+  echo "The user does not have the mysql cli installed or it is not available on the PATH."
+else
+  echo "The user has the mysql cli available. Use it to work with mysql."
+  echo "Always execute mysql statements non-interactively from the CLI using one of two options:"
+  echo '`-e "${SQL_QUERY}"` for single sql statements'
```
cloudnautique, line 14: these backticks need to be escaped, the shell is trying to execute them, this successfully runs on my mac.
```#!/bin/bash

if ! command -v mysql; then
  echo "The user does not have the mysql cli installed or it is not available on the PATH."
else
  mysql_cmd=$(command -v mysql)
  echo "The user has the mysql cli available. Use it to work with mysql."
  echo "Always execute mysql statements non-interactively from the CLI using one of two options:"
  echo '`-e "${SQL_QUERY}"` for single sql statements'
  echo "\`< \/path/to/script.sql\` for longer scripts"
  echo ""
  echo 'The password is pre-set using the `$MYSQL_PWD` environment variable'

  echo "the following text is the help output of the mysql cli:"
  echo "\`\`\`"
  ${mysql_cmd} --help
  echo "\`\`\`"
fi
```

PR #15, opened 2024-08-13
chore: initial hubspot crm tools

drpebcak:


apis/hubspot/crm/companies/context.gpt
```diff
@@ -0,0 +1,5 @@
+name: context
+tools: sys.http.html2text
```
g-linville, line 2: Does this need to be `share tools` instead? I forget but want to make sure.

g-linville (approved):

apis/hubspot/crm/companies/context.gpt
drpebcak: I don't think so because I just want this context tool to use it, I don't want to give that to the end user

PR #24, opened 2024-08-21
chore: add current time context

thedadams:


apis/outlook/calendar/manage/tool.gpt
```diff
@@ -1,4 +1,6 @@
 Name: Manage Outlook Calendar
 Description: Provides access to the Microsoft Outlook Calendar API
+Share Context: github.com/gptscript-ai/context/current-time
 Share Tools: openapi.yaml
 Share Credential: github.com/gptscript-ai/gateway-oauth2 as outlook.calendar.manage with GPTSCRIPT_GRAPH_MICROSOFT_COM_BEARER_TOKEN as env and microsoft365 as integration and "Calendars.Read Calendars.Read.Shared Calendars.ReadWrite Group.Read.All GroupMember.Read.All User.Read offline_access" as scope
+Type: context
```
g-linville, line 6: Why are these all type `context`? I designed them to be included like `tools: github.com/gptscript-ai/tools/apis/outlook...`

g-linville: Actually I'm not sure whether it makes a functional difference if we do `tools:` or `context:` for this...

thedadams: So they can `Share Context`. It is my understanding that only context tools can do that.

thedadams: >Actually I'm not sure whether it makes a functional difference if we do tools: or context: for this...

This is my understand as well since both can do `Share Tools`.

g-linville (approved):

PR #32, opened 2024-08-27
enhance: add Outlook Mail code tools

g-linville:
Much like the Outlook Calendar code tools, these aren't being used anywhere yet. I want to get them merged first, and then switch the existing `manage` and `read` tools over to using these.


apis/outlook/mail/code/pkg/global/global.go
```diff
@@ -0,0 +1,8 @@
+package global
+
+const CredentialEnv = "GPTSCRIPT_GRAPH_MICROSOFT_COM_BEARER_TOKEN"
+
+var (
+	ReadOnlyScopes = []string{"TODO"}
+	AllScopes      = []string{"TODO"}
```
thedadams, line 7: Do these need to be replaced with something else? If not, can we add a comment explaining?

thedadams (approved):

apis/outlook/mail/code/pkg/global/global.go
g-linville: Whoops lol. Yeah this proves that we probably don't need to specify scopes. I think I'll add them in anyway, in case we for some reason change how we handle authn/z.

g-linville: Fixed

apis/outlook/mail/code/tool.gpt
```diff
@@ -0,0 +1,59 @@
+Name: listMailFolders
+Description: Lists all available mail folders.
+
+#!${GPTSCRIPT_TOOL_DIR}/bin/gptscript-go-tool listMailFolders
+
+---
+Name: listMessages
```
cjellick, line 7: why do you use the word message instead of email throug out this?

g-linville: An individual email is referred to as a message in the Graph API, which is why I stuck with that naming. Though I could see how it could be less helpful to the LLM. I can just update the naming within the tool.gpt file if you want, since that's all the LLM will see.

njhale (approved):

PR #44, opened 2024-09-03
enhance: remove list issues and prs commands

njhale:
Issue search can be used to list issues and PRs with better results than
the basic list commands. One of the core problems is that these list
commands only return open issues and PRs. Remove them in favor of
relying on issue search.

### DIFF
--- a/apis/github/code/src/tools.js
+++ b/apis/github/code/src/tools.js
@@ -19,32 +19,6 @@ export async function listRepos(octokit, owner) {
     });
 }
 
-export async function listIssues(octokit, owner, repo) {
-    const issues = await octokit.issues.listForRepo({
-        owner,
-        repo,
-        state: 'open',
-        per_page: 100
-    });
-
-    issues.data.forEach(issue => {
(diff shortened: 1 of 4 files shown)


apis/github/code/src/tools.js
```diff
@@ -79,13 +53,34 @@ export async function deleteIssue(octokit, owner, repo, issueNumber) {
...
+        const { data: { type } } = await octokit.users.getByUsername({ username: owner });
+
+        // Construct the query based on the presence of repo and owner type
+        const ownerQualifier = type === 'User' ? `user:${owner}` : `org:${owner}`;
+        q = repo ? `repo:${owner}/${repo}` : ownerQualifier;
+    } else if (repo) {
+        // If no owner but a repo is given, search within that repo (this case is unlikely, so you might consider erroring out)
+        throw new Error('Repository given without an owner. Please provide an owner.');
+    } else {
+        // If neither owner nor repo is provided, throw an error or perform a global search
```
g-linville, line 70: Seems like these two comments should be removed

g-linville (approved):

PR #45, opened 2024-09-03
chore: initial split of read/write tools

drpebcak:


apis/hubspot/crm/write/tool.gpt
```diff
@@ -0,0 +1,41 @@
+Name: HubSpot CRM Write
+Type: context
+Description: Read and Write Tools for working with HubSpot CRM
+Share Tools: ./hubspot-crm-write.yaml, ../read
```
g-linville, line 4: Rather than sharing the read tool from the other directory, I would recommend sharing the OpenAPI file here.

drpebcak: I don't just need the openapi file though, i need the context that is in the tool.gpt and the helpers tools as well.

thedadams (approved):

PR #52, opened 2024-09-10
chore: convert some tools to code

drpebcak:


apis/hubspot/crm/helpers/createAssociation.py
```diff
@@ -0,0 +1,49 @@
...
+        "associationTypeId": resp.results[0].type_id,
+    }]
+
+    try:
+        client.crm.associations.v4.basic_api.create(object_type=from_object_type, object_id=from_object_id,
+                                                    to_object_type=to_object_type, to_object_id=to_object_id,
+                                                    association_spec=association_spec)
+    except ApiException as e:
+        print("Exception when calling associations->create: %s\n" % e)
+        print("Try mapping the objects the other way around.")
```
g-linville, line 26: nit: I think this should be rephrased to something like this:

```suggestion
        print("Attempting to map the objects the other way around...")
```

apis/hubspot/crm/helpers/getMe.py
```diff
@@ -0,0 +1,21 @@
...
+token = os.getenv("GPTSCRIPT_API_HUBAPI_COM_BEARER_TOKEN")
+client = HubSpot(access_token=token)
+
+
+def get_me():
+    try:
+        user = client.auth.oauth.access_tokens_api.get(token)
+        print(f"Current user is {user.user}. UserId is {user.user_id}.")
+    except Exception as e:
+        print(f"Error getting current user information: {e}")
```
g-linville, line 14: nit: I think we should exit(1) after this

```suggestion
    except Exception as e:
        print(f"Error getting current user information: {e}")
        exit(1)
```

g-linville (approved):

cjellick (approved):

PR #60, opened 2024-09-25
enhance: add more notion tools

g-linville:
This PR adds four new tools for Notion, mostly to support writing to databases. I also introduced various other improvements to the code, including removing the `share credential` lines and instead defining the credential in each individual tool.

### DIFF
--- a/apis/notion/code/context.gpt
+++ b/apis/notion/code/context.gpt
@@ -1,0 +1,18 @@
+name: addDatabaseRowInstructions
+type: context
+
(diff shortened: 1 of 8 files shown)


drpebcak (approved):

g-linville:
Re-requesting review because I pushed another commit to fix the read-only operations so that they use the correct credential.

drpebcak (approved):

StrongMonkey (approved):

PR #62, opened 2024-09-27
chore: improve behaviour around inserting cells with multiline content

drpebcak:


g-linville (approved):
I appreciate the British spelling of behaviour

PR #63, opened 2024-09-30
feat: add excel tools

g-linville:


drpebcak (approved):
I know we don't want to nest these too far, but I wonder if it might be time to drop these under a 'microsoft365' folder or something like that, to make it clear that these tools for excel, outlook, etc are for the m365 versions.

StrongMonkey:
>I know we don't want to nest these too far, but I wonder if it might be time to drop these under a 'microsoft365' folder or something like that, to make it clear that these tools for excel, outlook, etc are for the m365 versions.

Good point, or something like `o365-excel`.

StrongMonkey (approved):

g-linville:
I think a reorganization of the folders should be discussed with the team. For now, I'm going to merge this.

PR #78, opened 2024-10-08
enhance: make slack tools self-contained

njhale:
- Share slack tool dependencies to make tools "self-contained"; i.e. Each tool can be imported and used without manually importing other tools
- Add tool "bundle" for slack

### DIFF
--- a/slack/tool.gpt
+++ b/slack/tool.gpt
@@ -1,3 +1,9 @@
+---
+Name: Slack
+Description: Tools for interacting with Slack
+Share Tools: List Channels, Search Channels, Get Channel History, Get Channel History by Time, Get Thread History, Search Messages, Send Message, Send Message in Thread, List Users, Search Users, Send DM, Send DM in Thread, Get Message Link, Get DM History, Get DM Thread History
+
+---
 Name: List Channels
 Description: List all channels in the Slack workspace. Returns the name and ID for each channel
 Share Context: Slack Context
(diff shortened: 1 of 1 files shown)


apis/slack/tool.gpt
```diff
@@ -19,6 +19,7 @@ Name: Get Channel History
 Description: Get the chat history for a channel in the Slack workspace
 Share Context: Slack Context
 Credential: github.com/gptscript-ai/gateway-oauth2 as slack.write with SLACK_TOKEN as env and slack as integration and "channels:history groups:history im:history mpim:history channels:read files:read im:read search:read team:read users:read groups:read chat:write groups:write mpim:write im:write" as scope
+Share Tools: List Channels
```
g-linville, line 22: In each of the places where we share `List Channels`, can we also share `Search Channels`? The latter is a lot more useful for Slack workspaces that have a ton of channels.

apis/slack/tool.gpt
```diff
@@ -97,6 +102,7 @@ Name: Send DM
 Description: Send a direct message to a user.
 Share Context: Slack Context
 Credential: github.com/gptscript-ai/gateway-oauth2 as slack.write with SLACK_TOKEN as env and slack as integration and "channels:history groups:history im:history mpim:history channels:read files:read im:read search:read team:read users:read groups:read chat:write groups:write mpim:write im:write" as scope
+Share Tools: List Users
```
g-linville, line 105: Same here, we should include `Search Users`.

g-linville (changes requested):
One other thing - I've found that all of the other list and search commands are useful to include with `Search Messages`, so that the LLM can include modifiers like `in:` and `from:`. Since they aren't strictly required though, I'm not really sure what to do...

slack/tool.gpt
```diff
@@ -1,3 +1,9 @@
+---
+Name: Slack
+Description: Tools for interacting with Slack
+Share Tools: List Channels, Get Channel History, Get Channel History by Time, Get Thread History, Search Messages, Send Message, Send Message in Thread, List Users, Search Users, Send DM, Send DM in Thread, Get Message Link, Get DM History, Get DM Thread History
```
g-linville, line 4: `Search Channels` is missing from here.

g-linville (changes requested):

g-linville (approved):

PR #81, opened 2024-10-09
enhance: add youtube bundle tool and metadata

njhale:
- Rename tools and descriptions to more accurately convey their purpose
- Attempt to generate a transcript when one isn't available from YouTube
- Add tool and associated metadata to bundle all YouTube related tools

### DIFF
--- a/google/youtube/tool.gpt
+++ b/google/youtube/tool.gpt
@@ -1,24 +1,35 @@
-Name: Transcribe Youtube Video
+---
+Name: YouTube
+Metadata: bundle: true
+Description: Tools for interacting with YouTube
+Share Tools: Transcribe YouTube Video
+
+---
+Name: Transcribe YouTube Video
 Description: Get a text transcript of a YouTube video and clean it up.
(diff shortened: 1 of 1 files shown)


google/youtube/tool.gpt
```diff
@@ -1,3 +1,9 @@
+---
+Name: YouTube
+Description: Tools for interacting with YouTube
+Share Tools: Transcribe Youtube Video, Youtube Transcriber, Advanced Youtube Transcriber
```
njhale, line 4: Not sure folks will use `Youtube Transcriber` on its own, so maybe it doesn't make sense to include it in the bundle.

njhale: I've played around with this to make all the tools useful in some capacity.

google/youtube/tool.gpt
```diff
@@ -1,24 +1,35 @@
-Name: Transcribe Youtube Video
+---
+Name: YouTube
+Metadata: bundle: true
+Description: Tools for interacting with YouTube
+Share Tools: Transcribe YouTube Video
```
njhale, line 5: We're omitting the two "helper" tools to simplify UX.

g-linville (approved):

PR #82, opened 2024-10-09
enhance: make notion tools self-contained

njhale:
- Make notion tool self-contained s.t. the necessary tools to
  discover required parameters are shared
- Add the notion tool bundle

### DIFF
--- a/notion/tool.gpt
+++ b/notion/tool.gpt
@@ -1,6 +1,13 @@
+---
+Name: Notion
+Description: Tools for interacting with Notion
+Share Tools: Get Page, Get Database, Search, Create Page, List Users, Get Database Properties, Add Database Row, Update Database Row
+
+---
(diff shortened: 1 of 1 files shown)


notion/tool.gpt
```diff
@@ -62,12 +73,16 @@ Name: Update Database Row
 Description: Updates an existing row in a database in Notion. Call Get Database Properties prior to this to get the properties of the database.
 Credential: github.com/gptscript-ai/gateway-oauth2 as notion.write with NOTION_TOKEN as env and notion as integration
 Share Context: addDatabaseRowInstructions from context.gpt
+Share Tools: Search
```
g-linville, line 76: This one probably also needs `Get Database`, so that it knows which rows there are.

g-linville (approved):

PR #88, opened 2024-10-10
add: knowledge demo documentloaders

iwilltry42:


iwilltry42:
I guess we won't go this route

PR #100, opened 2024-10-15
feat: update tools to use datasets

g-linville:
This PR updates several of our tools to use the new datasets feature. A dataset is basically a set of files that get stored in the workspace, where each file contains one element of arbitrary data.

We might want to adjust this, but I implemented this only for tools that return lists that are typically large, and we only create the dataset if there are more than 10 elements in the list (less than 10, and we just print it out like usual).


drpebcak (approved):

thedadams (approved):

g-linville:
Re-requesting review because I made changes so that we just print the results if there are failures with the dataset. This is for backwards compatibility with older versions of GPTScript that do not have datasets in the SDK yet.

outlook/calendar/pkg/printers/events.go
```diff
@@ -12,6 +12,28 @@ import (
...
+			calendarName = groupName
+		}
+	}
+
+	var sb strings.Builder
+	sb.WriteString("Subject: " + util.Deref(event.GetSubject()) + "\n")
+	sb.WriteString("  ID: " + util.Deref(event.GetId()) + "\n")
+	sb.WriteString("  Start: " + util.Deref(event.GetStart().GetDateTime()) + "\n")
+	sb.WriteString("  End: " + util.Deref(event.GetEnd().GetDateTime()) + "\n")
+	sb.WriteString("  In calendar: " + calendarName + " (ID " + calendar.ID + ")\n")
```
njhale, line 33: nit: I know it saves tokens to print it this way, but at some point I think minified JSON streams with `Share Context` containing a JSON schema of the output (along with descriptions) would be both more maintainable and give the LLM a better idea of the output, especially when we're dealing with structured data.

github/src/tools.js
```diff
@@ -143,6 +186,29 @@ export async function listPRComments(octokit, owner, repo, prNumber) {
...
+    if (data.length > 10) {
+        try {
+            const gptscriptClient = new GPTScript();
+            const dataset = await gptscriptClient.createDataset(
+                process.env.GPTSCRIPT_WORKSPACE_DIR,
+                `${owner}_${repo}_pr_${prNumber}_comments`,
+                `Comments for PR #${prNumber} in ${owner}/${repo}`
+            );
+            for (const comment of data) {
+                await gptscriptClient.addDatasetElement(
```
njhale, line 199: nit: would be convenient to be able to initialize a dataset with a set of elements and/or add N elements to a dataset in a single statement?

njhale (approved):

outlook/calendar/pkg/util/util.go
```diff
@@ -18,3 +18,20 @@ func Map[T, U any](arr []T, f func(T) U) []U {
...
+func Flatten[T any](arr [][]T) []T {
+	var out []T
+	for _, v := range arr {
+		out = append(out, v...)
+	}
+	return out
+}
+
+func MapValues[T comparable, U any](m map[T]U) []U {
+	// The maps.Values function returns an iterator, but we want a slice.
```
thedadams, line 31: It doesn't seem like the implementation needs a slice, but I'm OK with not introducing iterators at this point.

thedadams (approved):

github/src/tools.js
g-linville: I agree that this would be good. It will take some time for me to add that, but that's an improvement I can make while I wait for further feedback on datasets. I'll do this in a follow-up PR.

outlook/calendar/pkg/printers/events.go
g-linville: I think I'll also do this in a follow-up PR. This is how we've been giving the data to the LLM and it has worked well enough for now. I want to get this in so that Darren can try it out and tell me what needs to change.

PR #102, opened 2024-10-17
fix: make google search refinement more conservative 

njhale:
Make Google Search query refinement more conservative by adding the `Current Date and Time` context tool to internal tool calls and tweaking the refinement tool prompt. This should help prevent the refined query changing too much and negatively impacting the results, especially when the query involves dates.

e.g. before this change "Who won the 2024 Super Bowl?" is refined to "2024 Super Bow predictions and favorites" which results in content from old articles being returned. After adding this change, the query is refined to "Who won the Super Bowl 2024" which results in content that can actually answer the question.

This change also suppresses `npm run` script body printing to ensure the tool produces valid JSON.

### DIFF
--- a/google/search/src/refineQuery.ts
+++ b/google/search/src/refineQuery.ts
@@ -5,7 +5,7 @@ export async function refineQuery(query: string): Promise<string> {
     agents: [],
     arguments: { type: 'object' },
     chat: false,
-    context: [],
+    context: ['Current Date and Time from github.com/otto8-ai/tools/time'],
     credentials: [],
     description: '',
     export: [],
@@ -19,8 +19,8 @@ export async function refineQuery(query: string): Promise<string> {
     tools: [],
     modelName: 'gpt-4o-mini',
     instructions: `
-    Refine the query below to improve its Google Search results.
-    The refined query should preserve the inferred intent of the original query.
+    Refine the query below to improve the quality of its Google Search results if necessary.
+    Be conservative with refinements so that the original meaning is more or less preserved.
     Do not quote the output.
 
     Query: ${query}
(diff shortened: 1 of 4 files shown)


google/search/src/getMarkdown.ts
```diff
@@ -11,9 +11,7 @@ import * as cheerio from 'cheerio'
 export async function getMarkdown(page: Page, url: string): Promise<string> {
   try {
     await page.goto(url, { timeout: 1000 })
-  } catch (e) {
-    console.warn('slow page:', url)
```
njhale, line 15: Looking for some guidance here. I removed these log statements to ensure the tool returned only JSON on the happy path. It was nice to know when a page was slow or couldn't be loaded, but it doesn't stop the tool from working.

g-linville: I assume `console.warn` prints to stdout. Does `console.error` work here? Iirc, gptscript does not send stderr to the LLM when the tool call is finished, only stdout.

g-linville (approved):

google/search/src/getMarkdown.ts
njhale: > I assume console.warn prints to stdout.

Actually, it's an alias for console.error, which already prints to stderr

>  Iirc, gptscript does not send stderr to the LLM when the tool call is finished, only stdout.

just double checked in otto and you're right. I must have misread the output when I initially tested the same query with `gptscript`, so these warn statements can stay.

njhale: okay, updated

thedadams (approved):

PR #105, opened 2024-10-17
feat: add vision and image generation tools

njhale:
Add an `Images` tool bundle with tools for image generation and analysis.

Note: Open to suggestions on better names for this bundle.

### DIFF
--- a/images/src/analyze.ts
+++ b/images/src/analyze.ts
@@ -1,0 +1,70 @@
+import { fileTypeFromBuffer } from 'file-type';
+import fs from 'fs';
+import OpenAI from 'openai';
+import { ChatCompletionContentPartImage } from 'openai/resources/chat/completions';
+
+const resolveImageURL = async (image: string): Promise<string> => {
+  const uri = new URL(image)
+  switch (uri.protocol) {
+    case 'http:':
+    case 'https:':
+      return image;
+    case 'file:': {
+      const filePath = image.slice(7);
+      const data = fs.readFileSync(filePath);
+      const mime = (await fileTypeFromBuffer(data))?.mime;
+      if (mime !== 'image/jpeg' && mime !== 'image/png') {
+        throw new Error('Unsupported mimetype');
+      }
+      const base64 = data.toString('base64')
+      return `data:${mime};base64,${base64}`
+    }
+    default:
+      throw new Error('Unsupported protocol')
+  }
+};
+
+const analyzeImages = async (
+  model: string = 'gpt-4o',
+  prompt: string = '',
+  images: string[] = [],
+): Promise<void> => {
+  if (!prompt) {
+    throw new Error('No prompt provided. Please provide a prompt to send to the vision model.');
+  }
+  if (images.length === 0) {
+    throw new Error('No images provided. Please provide a list of images to send to the vision model.');
+  }
+
+  const content = await Promise.all(
+    images.map(async image => ({
+      type: 'image_url',
+      image_url: {
+        detail: 'auto',
+        url: await resolveImageURL(image),
+      },
+    }))
+  ) as ChatCompletionContentPartImage[];
+
+  const openai = new OpenAI();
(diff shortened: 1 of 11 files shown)


images/src/tools.ts
```diff
@@ -0,0 +1,39 @@
...
+                process.env.PROMPT,
+                // Split the string into an array of image URLs, while being careful to handle URLs that contain commas.
+                process.env.IMAGES?.split(/(?<!\\),/).map(image => image.replace(/\\,/g, ',')),
+                parseInt(process.env.MAX_TOKENS ?? '2048'),
+            )
+            break
+        case 'generateImages':
+            generateImages(
+                process.env.PROMPT,
+                process.env.MODEL,
```
g-linville, line 24: It doesn't look this this is exposed as a `Param:` for this tool. Should we add it?

images/src/generate.ts
```diff
@@ -0,0 +1,39 @@
+import OpenAI from 'openai';
+
+type ImageSize = '1024x1024' | '256x256' | '512x512' | '1792x1024' | '1024x1792';
```
g-linville, line 3: If these are the only size options, I think we should do more validation around this and also let the model know in the parameter description that it has to be one of these sizes.

images/src/generate.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+
+  const openai = new OpenAI();
+
+  try {
+    const response = await openai.images.generate({
+      model,
+      prompt,
+      size: size as ImageSize,
```
g-linville, line 24: (Specifically here is where I think we need to do validation, unless this throws an exception if `size` has no corresponding `ImageSize`. I don't know if that's what happens.)

images/src/tools.ts
njhale: Not sure users should be able to change the model yet since some models don't support vision.

drpebcak: There's decent vision support out there, its really image-generation that's lacking.

Without a param for model, this is a regression for our support of vision with the ollama-provider.

images/src/analyze.ts
```diff
@@ -0,0 +1,71 @@
...
+      image_url: {
+        detail: 'auto',
+        url: await resolveImageURL(image),
+      },
+    }))
+  ) as ChatCompletionContentPartImage[];
+
+  const openai = new OpenAI();
+  const response = await openai.chat.completions.create({
+    model: 'gpt-4o',
```
drpebcak, line 51: Not sure we want this hardcoded. There are other models that can support vision if we are more flexible.

drpebcak (changes requested):

images/src/tools.ts
njhale: Got it. I'll add the model param back to the tool.

images/src/generate.ts
njhale: done

images/src/generate.ts
njhale: done

images/src/tools.ts
njhale: done

images/src/analyze.ts
njhale: I plumbed this down from the new param.

thedadams (approved):
This looks good. However, I do remember a request to write tools in Python rather than Javascript. Not sure if that makes sense here.

njhale:
> This looks good. However, I do remember a request to write tools in Python rather than Javascript. Not sure if that makes sense here.

@thedadams I had a conversation with Craig about this right before he went on vacation. The gist of it was to use whatever language makes the most sense for the tool we're working on. In this case, I felt like typescript was a good fit because other tools in this repo use it and vision was already a nodejs app.

I can convert this to python if you feel strongly about it.

thedadams:
@njhale Doesn't matter to me.

drpebcak (approved):

PR #106, opened 2024-10-18
feat: add word tools

njhale:
Add tools to list and read Microsoft Word documents using the Graph API.


word/pkg/client/client.go
```diff
@@ -0,0 +1,27 @@
...
+	"os"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	msgraphsdkgo "github.com/microsoftgraph/msgraph-sdk-go"
+
+	"github.com/gptscript-ai/tools/word/pkg/global"
+)
+
+// StaticTokenCredential is taken from https://github.com/gptscript-ai/mail-assistant/blob/10944805801bbb6f71eccefd1bea5f114fded164/pkg/mstoken/auth.go
```
thedadams, line 14: nit: remove comment.
Opinion: it's not relevant where this is copied from, just that it works.

thedadams (approved):

word/pkg/client/client.go
njhale: done

word/tool.gpt
```diff
@@ -0,0 +1,43 @@
...
+
+#!${GPTSCRIPT_TOOL_DIR}/bin/gptscript-go-tool listDocs
+
+---
+Name: Get Doc
+Description: Gets the contents of a word document
+Share Context: Word Context
+Credential: ./credential
+Share Tools: List Docs
+Param: doc_id: ID word document to get
```
g-linville, line 21: nit:

```suggestion
Param: doc_id: ID of the word document to get
```

g-linville (approved):

word/tool.gpt
```diff
@@ -0,0 +1,43 @@
+---
+Name: Word
+Description: Tools for interacting with Microsoft Word documents
+Metadata: bundle: true
+Share Tools: List Docs, Get Doc
+
+---
+Name: List Docs
+Description: List all word documents available to the user
```
drpebcak, line 9: nit - i believe capitalization matters to the LLM to some extent, so all usages of `Word` as in `MS Word` should have the same capitalization:

```suggestion
Description: List all Word documents available to the user
```

drpebcak (changes requested):

word/tool.gpt
njhale: Thanks for catching that! Fixed.

word/tool.gpt
njhale: I changed the capitalization and used `Microsoft Word` more ubiquitously.

Only question is whether the name of the category/bundle should change from `Word` to `Microsoft Word`, and in that case should I restructure the directories to align with the convention we're using for google products; e.g. `{word|excel|outlook}` -> `microsoft/{word|excel|outlook}`?

I know Darren was looking to keep the directory structure as flat as possible.

drpebcak: i dont know if darren would agree, but i think that the directory structure should match googles, and it should be something like `microsoft365` to reflect the fact that these tools are for m365 only, not local documents. i wouldnt make any changes on that without passing it by darren or craig. at the very least though we should standardize, so either this or google should be changed to match at some point. i wouldnt let it hold up this pr.

drpebcak (approved):

word/tool.gpt
njhale: cool, I'll merge this then follow up in slack with our discussion

PR #114, opened 2024-10-23
feat: add browser tools

njhale:
Add MVP `Browser` tools with updated session management for use in Otto.

This means:
- The `Browser` tool daemon doesn't store session data in the workspace directory
-- which is now deleted after each tool call -- and instead creates an independent session directory and browser process for each unique workspace ID it encounters 
- The session directories and browser processes are still ephemeral, but now have a 10 minute TTL from last use
- Workspace IDs are taken from the headers of requests to the tool daemon (can't use the environment variable anymore because tool daemons are effectively singletons in Otto)

### DIFF
--- a/browser/src/server.ts
+++ b/browser/src/server.ts
@@ -1,0 +1,153 @@
+import express, { type Request, type Response } from 'express'
+import bodyParser from 'body-parser'
+import { type Page } from 'playwright'
+import { browse, filterContent } from './browse.ts'
+import { fill } from './fill.ts'
+import { enter } from './enter.ts'
+import { scrollToBottom } from './scrollToBottom.ts'
+import { randomBytes } from 'node:crypto'
+import { getSessionId, SessionManager } from './session.ts'
+
+async function main (): Promise<void> {
+  console.log('Starting browser server')
+
+  const app = express()
+  // Get port from the environment variable or use 9888 if it is not defined
+  const port = process.env.PORT ?? 9888
+  delete (process.env.GPTSCRIPT_INPUT)
+  app.use(bodyParser.json())
+
+  const sessionManager = await SessionManager.create()
+
+  // gptscript requires "GET /" to return 200 status code
+  app.get('/', (_req: Request, res: Response) => {
+    res.send('OK')
+  })
+
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  app.post('/*', async (req: Request, res: Response) => {
+    const data = req.body
+
+    const model: string = data.model ?? 'gpt-4o-mini'
+    const website: string = data.website ?? ''
+    const userInput: string = data.userInput ?? ''
+    const keywords: string[] = (data.keywords ?? '').split(',')
+    const filter: string = data.filter ?? ''
+
+    try {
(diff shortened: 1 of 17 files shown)


browser/tool.gpt
```diff
@@ -0,0 +1,122 @@
...
+
+---
+Name: Forward
+Share Context: Browser Context
+Tools: service
+Description: Navigates to forward to the next site in history.
+Params: tabID: (required) The ID of the tab.
+Tools: service
+
+#!http://service.daemon.gptscript.local/forward
```
njhale, line 90: A lot of the tools here aren't useful on their own unless they share pretty much all of the other tools.

What would folks think if I tag everything but the bundle and `Get Page Content` with `Metadata: index: false`?

browser/src/session.ts
```diff
@@ -0,0 +1,130 @@
...
+export class SessionManager {
+  private readonly sessions = new Map<string, ManagedSession>()
+  private readonly sessionsLock: Mutex = new Mutex()
+
+  private constructor () {
+  }
+
+  static async create (): Promise<SessionManager> {
+    sessionManager ??= new SessionManager()
+    return sessionManager
```
njhale, line 40: Session directories can become orphaned if a daemon with "living" sessions is killed. Should we delete orphaned directories at when the session manager is created to handle that?

browser/tool.gpt
thedadams: I think that makes sense.

thedadams (approved):

browser/src/session.ts
g-linville: I think this is a good idea. Orphaned session directories could end up taking up a lot of drive space.

g-linville (approved):

browser/src/server.ts
```diff
@@ -0,0 +1,158 @@
...
+import { randomBytes } from 'node:crypto'
+import { getSessionId, SessionManager } from './session.ts'
+
+async function main (): Promise<void> {
+  console.log('Starting browser server')
+
+  const app = express()
+  // Get port from the environment variable or use 9888 if it is not defined
+  const port = process.env.PORT ?? 9888
+  delete (process.env.GPTSCRIPT_INPUT)
```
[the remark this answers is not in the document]
g-linville: I think I may have been the one to add this in the old browser tool repo. I was seeing cases where GPTSCRIPT_INPUT would be passed to internal tool calls and possibly mess them up, if they were trying to read their own args from that env var. At least, I think that's why I was deleting this here.

browser/src/session.ts
njhale: done

browser/src/server.ts
```diff
@@ -0,0 +1,158 @@
...
+        } else {
+          page = await browserContext.newPage()
+          openPages.set(tabID, page)
+        }
+        await page.bringToFront()
+
+        let allElements = false
+        if (data.allElements === 'true' || data.allElements === true) {
+          allElements = true
+        }
```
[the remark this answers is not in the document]
njhale: `allElements` wasn't actually being used anymore. Removed.

browser/tool.gpt
njhale: done

PR #117, opened 2024-10-24
fix: trim space on data source urls

thedadams:


knowledge/data-sources/website/main.go
```diff
@@ -100,6 +101,14 @@ func main() {
 		metadata.Output.State.WebsiteCrawlingState.Pages = make(map[string]PageDetails)
 	}
 
+	for i := range metadata.Input.WebsiteCrawlingConfig.URLs {
+		metadata.Input.WebsiteCrawlingConfig.URLs[i] = strings.TrimSpace(metadata.Input.WebsiteCrawlingConfig.URLs[i])
+	}
+
+	for i := range metadata.Input.Exclude {
```
StrongMonkey, line 108: Exclude can be removed, it is a legacy field that is not used anymore.

StrongMonkey (approved):

njhale (approved):

PR #119, opened 2024-10-24
enhance: add default timezone tool to outlook calendar

njhale:
Add the `Get Default Timezone` tool to `Outlook Calendar` to avoid prompting
the user for a timezone to use when creating or modifying events.

The tool returns the timezone from their Outlook mailbox settings.

Addresses: https://github.com/otto8-ai/otto8/issues/169

~Note: Some permissions need to be approved [here](https://portal.azure.com/#view/Microsoft_AAD_RegisteredApps/ApplicationMenuBlade/~/CallAnAPI/appId/b89e586d-26d5-4126-8012-dbad9c137417/defaultBlade/CallAnAPI) before I can test this (even though that page seems to say otherwise, the challenge page is still blocking me on admin approval for app permissions cc @drpebcak)~

### DIFF
--- a/outlook/calendar/credential/tool.gpt
+++ b/outlook/calendar/credential/tool.gpt
@@ -9,5 +9,7 @@ Share Credential: ../../../oauth2 as outlook.calendar.write
             Group.Read.All
             Group.ReadWrite.All
             GroupMember.Read.All
-            User.Read offline_access" as scope
+            User.Read
+            MailboxSettings.Read
+            offline_access" as scope
 Type: credential
--- a/outlook/calendar/main.go
+++ b/outlook/calendar/main.go
@@ -110,6 +110,11 @@ func main() {
 			fmt.Println(err)
 			os.Exit(1)
 		}
+	case "getDefaultTimezone":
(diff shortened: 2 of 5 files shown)


drpebcak:
I've clicked the button, so you should be able to test now.

outlook/calendar/credential/tool.gpt
```diff
@@ -9,5 +9,6 @@ Share Credential: ../../../oauth2 as outlook.calendar.write
             Group.Read.All
             Group.ReadWrite.All
             GroupMember.Read.All
-            User.Read offline_access" as scope
+            User.Read
+            MailboxSettings.Read offline_access" as scope
```
drpebcak, line 13: small nit, but why do we have `offline_access` on the same line as another permission here when all the others have newlines?

njhale: Good question. I must have missed that when I first added this credential tool.

will fix

njhale: fixed

StrongMonkey (approved):

drpebcak (approved):

PR #120, opened 2024-10-24
chore: go tools: use AddDatasetElements

g-linville:
This should improve performance, since we're doing just one request and one tool call to add all the elements, instead of one per element.


outlook/calendar/pkg/commands/listEvents.go
```diff
@@ -44,35 +43,42 @@ func ListEvents(ctx context.Context, start, end time.Time) error {
...
+			return fmt.Errorf("failed to create dataset: %w", err)
+		}
 
-						return fmt.Errorf("failed to add event to dataset: %w", err)
-					}
+		names := map[string]any{} // Store names to avoid duplicates
+		var elements []gptscript.DatasetElement
+		for cal, events := range calendarEvents {
+			for _, event := range events {
+				name := util.Deref(event.GetSubject()) + "_" + util.Deref(cal.Calendar.GetName()) + "_" + util.Deref(event.GetStart().GetDateTime())
```
njhale, line 61: I just checked and it is possible to create more than one distinct event in a calendar with the same subject and start time (and end time, content, etc too). I think we need to keep all the events, regardless of whether they share the same user-provided fields. There should be a unique ID field on these objects we can add to the name to do that.

g-linville: Oh, thanks. I'll fix this.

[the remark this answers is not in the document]
g-linville: fixed

outlook/calendar/pkg/commands/listEvents.go
```diff
@@ -44,35 +43,42 @@ func ListEvents(ctx context.Context, start, end time.Time) error {
...
+					elements = append(elements, gptscript.DatasetElement{
+						DatasetElementMeta: gptscript.DatasetElementMeta{
+							Name:        name,
+							Description: util.Deref(event.GetBodyPreview()),
+						},
+						Contents: printers.EventToString(ctx, c, cal, event),
+					})
+					names[name] = struct{}{}
 				}
+
```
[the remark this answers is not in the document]
g-linville: this code has been removed

outlook/calendar/pkg/commands/listEvents.go
g-linville: fixed

outlook/mail/pkg/commands/listMailFolders.go
```diff
@@ -30,23 +30,35 @@ func ListMailFolders(ctx context.Context) error {
...
-				if err != nil {
-					return fmt.Errorf("failed to convert mail folder to string: %w", err)
-				}
-
-				if _, err = gptscriptClient.AddDatasetElement(ctx, os.Getenv("GPTSCRIPT_WORKSPACE_DIR"), dataset.ID, util.Deref(folder.GetId()), util.Deref(folder.GetDisplayName()), folderStr); err != nil {
-					return fmt.Errorf("failed to add element: %w", err)
-				}
+		workspaceID := os.Getenv("GPTSCRIPT_WORKSPACE_ID")
+
+		dataset, err := gptscriptClient.CreateDataset(ctx, workspaceID, "outlook_mail_folders", "Outlook mail folders")
```
thedadams, line 35: nitL is there a reason this one has `_` and some of the others don't?

thedadams (approved):

outlook/mail/pkg/commands/listMailFolders.go
g-linville: nope not really

outlook/mail/pkg/commands/listMessages.go
```diff
@@ -43,23 +43,35 @@ func ListMessages(ctx context.Context, folderID string) error {
...
-				messageStr, err := printers.MessageToString(message, false)
-				if err != nil {
-					return fmt.Errorf("failed to convert message to string: %w", err)
-				}
+		workspaceID := os.Getenv("GPTSCRIPT_WORKSPACE_ID")
 
-				if _, err = gptscriptClient.AddDatasetElement(ctx, os.Getenv("GPTSCRIPT_WORKSPACE_DIR"), dataset.ID, util.Deref(message.GetId()), util.Deref(message.GetSubject()), messageStr); err != nil {
-					return fmt.Errorf("failed to add element: %w", err)
-				}
+		dataset, err := gptscriptClient.CreateDataset(ctx, workspaceID, fmt.Sprintf("%s_outlook_mail", folderID), "Outlook mail messages in folder "+folderID)
```
drpebcak, line 48: noticed same thing here as @thedadams's comment, why `_` here but not other places?

drpebcak (approved):

PR #127, opened 2024-10-25
enhance: outlook calendar: support the creation of recurring meetings

g-linville:
Recurring meetings in the Outlook Calendar API are fairly complicated. To help handle this, I implemented an internal tool call to turn a plain English description of the recurrence logic into the proper fields required by the API. I have tested this thoroughly and it works well.

### DIFF
--- a/outlook/calendar/pkg/recurrence/recurrence_test.go
+++ b/outlook/calendar/pkg/recurrence/recurrence_test.go
@@ -187,11 +187,11 @@ func TestRecurrenceGeneration(t *testing.T) {
 		},
 		{
 			name:       "Relative Monthly Numbered",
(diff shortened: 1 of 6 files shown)


outlook/calendar/pkg/recurrence/recurrence_test.go
```diff
@@ -0,0 +1,317 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TODO - test these for real with Outlook
+func TestRecurrenceGeneration(t *testing.T) {
```
g-linville, line 11: I didn't set up any CI for these tests, but running them locally, they all pass. These test to make sure that the internal tool call to generate the recurrence object works the way we expect.

drpebcak (approved):

thedadams (approved):

PR #133, opened 2024-10-29
chore: make the data source credentials similar to other credentials

thedadams:
This change also fixes a small bug in the onedrive connector.

### DIFF
--- a/knowledge/data-sources/notion/credential/tool.gpt
+++ b/knowledge/data-sources/notion/credential/tool.gpt
@@ -1,0 +1,3 @@
+Name: Notion Data Source Credential
+Share Credential: ../../../../oauth2 as notion.sync-file with NOTION_TOKEN as env and notion as integration
+Type: credential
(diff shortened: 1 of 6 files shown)


knowledge/data-sources/notion/credential/tool.gpt
```diff
@@ -0,0 +1,3 @@
+Name: Notion Data Source Credential
+Share Credential: ../../../../oauth2 as notion.sync-file with NOTION_TOKEN as env and notion as integration
+Type: credential
```
njhale, line 3: nit: no newline at the end of this file.

njhale (approved):

knowledge/data-sources/notion/credential/tool.gpt
thedadams: Does this mean there is a new line and there shouldn't be? Or visa-versa?

njhale: there wasn't a new line but should be.

PR #137, opened 2024-10-29
enhance: add screenshot tool

cloudnautique:
Added a gptscript tool to take the screenshot of a tabID
Added API endopoint for tool to communicate with.
Added new method to take the screenshot and store in the workspace.

### DIFF
--- a/browser/package.json
+++ b/browser/package.json
@@ -17,7 +17,7 @@
     "typescript": "^5.3.3"
   },
   "dependencies": {
-    "@gptscript-ai/gptscript": "^0.9.5",
+    "@gptscript-ai/gptscript": "https://github.com/gptscript-ai/node-gptscript#306861b8b5aea2ea0481c180067a1687d2726530",
     "@types/turndown": "^5.0.4",
     "async-mutex": "^0.5.0",
     "body-parser": "^1.20.2",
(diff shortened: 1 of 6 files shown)


g-linville (approved):

browser/src/screenshot.ts
```diff
@@ -0,0 +1,20 @@
...
+export async function screenshot (page: Page): Promise<string> {
+  let screenshotName = `screenshot-${Date.now()}_${page.url().replace(/[^a-zA-Z0-9]/g, '_')}.png`
+  // Detect if we are running in otto8
+  if (process.env.OTTO_THREAD_ID !== undefined) {
+    screenshotName = `files/${screenshotName}`
+  }
+  const screenshot = await page.screenshot()
+
+  try {
+    await client.writeFileInWorkspace(screenshotName, screenshot)
```
njhale, line 15: this method will default to the workspace ID environment variable if it's not given as an argument, but since this runs as a daemon that won't change after startup. IIUC, this means that all users will be saving their screenshots to the same workspace.

Instead, I think the workspace ID should be extracted from the request headers, [like we do here](https://github.com/otto8-ai/tools/blob/f80a15a6bdf066994d20747261bd657866896381/browser/src/session.ts#L121).

browser/package.json
```diff
@@ -30,4 +30,4 @@
     "ts-node-dev": "^2.0.0",
     "turndown": "^7.1.3"
   }
-}
+}
```
njhale, line 33: nit: file is missing final newline

njhale (changes requested):

PR #138, opened 2024-10-29
fix: various fixes to Outlook Calendar tools

g-linville:
This PR:
- Improves the search tool by doing manual filtering on the subject and body preview
- Improves the prompting around creating events on group calendars
- Introduces the same ID translation strategy that we use in Outlook Mail, to help prevent the LLM from getting confused about event and calendar IDs


outlook/calendar/main.go
```diff
@@ -70,6 +70,11 @@ func main() {
 		if id := os.Getenv("CALENDAR_ID"); id != "" {
 			info.ID = id
 			info.Owner = graph.OwnerType(os.Getenv("OWNER_TYPE"))
+
+			if info.Owner == "" {
+				fmt.Println("Owner type is required")
+				os.Exit(1)
+			}
```
g-linville, line 77: The LLM was often forgetting to set this parameter. I improved the prompting in the parameter description, so it shouldn't forget as often, but if it does, we now return an error so that it retries.

outlook/calendar/pkg/commands/searchEvents.go
```diff
@@ -21,19 +22,30 @@ func SearchEvents(ctx context.Context, query string, start, end time.Time) error
...
+		}
+
+		for _, event := range result {
+			if strings.Contains(util.Deref(event.GetSubject()), query) {
+				calendarEventsInSubject[cal] = append(calendarEventsInSubject[cal], event)
+			} else if strings.Contains(util.Deref(event.GetBodyPreview()), query) {
+				calendarEventsInPreview[cal] = append(calendarEventsInPreview[cal], event)
+			}
 		}
 	}
```
g-linville, line 45: We used to search the events. This would basically list events but with a search parameter that didn't seem to work very well and sometimes returned syntax errors. At the end of the day, the LLM would get several seemingly irrelevant results.

With this change, we just list all the events within the time window, and then filter them ourselves by checking the subject and body preview (in a case-sensitive way) to see if they contain the query. Still not perfect, but it does better than what we had before.

outlook/calendar/tool.gpt
```diff
@@ -59,7 +59,7 @@ Param: start: (Required) The start time of the event, in RFC 3339 format.
 Param: end: (Required) The end time of the event, in RFC 3339 format. When scheduling a recurring event, this should be the end time of the first event in the series.
 Param: recurrence: (Optional) If the meeting should recur, describe in plain English how often it should occur (daily, weekly, monthly, yearly) and during which date range (first and last occurrence) or how many total times the event should occur. ALWAYS include the date of the first occurrence, and optionally the date of the last occurrence.
 Param: calendar_id: The unique ID of the calendar or group to add the event to. If unset, adds the event to the default calendar.
-Param: owner_type: The type of the owner of the calendar or group. Possible values are "user" or "group". Required if calendar_id is set.
+Param: owner_type: (Required if calendar_id is set) The type of the owner of the calendar or group. Possible values are "user" or "group".
```
njhale, line 62: nit: feels like we should be able to get by without `owner_type` if we already have a unique `calendar_id`, but I guess the SDK design doesn't make that easy. We could always inject the owner type into the ids we return from the list operation, but if the LLM can reliably figure this field out on it's own, that's probably unnecessary.

njhale (approved):

outlook/calendar/tool.gpt
g-linville: Yeah after I rearranged the prompt, the LLM seems to not have an issue figuring this out. Imo it's best to leave it a it is.

outlook/calendar/pkg/commands/searchEvents.go
```diff
@@ -21,19 +22,30 @@ func SearchEvents(ctx context.Context, query string, start, end time.Time) error
...
-		return fmt.Errorf("failed to search events: %w", err)
+		return fmt.Errorf("failed to list calendars: %w", err)
 	}
 
-	calendarEvents := map[graph.CalendarInfo][]models.Eventable{}
-	for cal, events := range result {
-		if len(events) > 0 {
-			calendarEvents[cal] = events
+	calendarEventsInSubject := map[graph.CalendarInfo][]models.Eventable{}
+	calendarEventsInPreview := map[graph.CalendarInfo][]models.Eventable{}
```
thedadams, line 31: nit: better to over allocate for bigger requests.
```suggestion
	calendarEventsInSubject := make(map[graph.CalendarInfo][]models.Eventable, len(calendars))
	calendarEventsInPreview := make(map[graph.CalendarInfo][]models.Eventable, len(calendars))
```

outlook/calendar/pkg/util/util.go
```diff
@@ -35,3 +35,26 @@ func MapValues[T comparable, U any](m map[T]U) []U {
 	}
 	return out
 }
+
+// Merge merges two maps where the values are arrays of the same type.
+// The resulting map will contain all keys from both maps, with each value containing the concatenated arrays from both maps.
+func Merge[T comparable, U any](one map[T][]U, two map[T][]U) map[T][]U {
+	out := make(map[T][]U)
+	checked := make(map[T]any)
```
thedadams, line 43: nit:
```suggestion
	out := make(map[T][]U, len(one) + len(two))
	checked := make(map[T]struct{})
```

thedadams (approved):

PR #145, opened 2024-10-30
chore: add more datasets to Slack

g-linville:


iwilltry42 (approved):

slack/src/tools.js
```diff
@@ -288,6 +302,18 @@ export async function getDMHistory(webClient, userIds, limit) {
...
         console.log('No messages found')
         return
+    } else if (history.messages.length > 10) {
+        try {
+            const gptscriptClient = new GPTScript()
+            const dataset = await gptscriptClient.createDataset(process.env.GPTSCRIPT_WORKSPACE_ID, `slack_dm_history_${userIds}`, `chat history for DM with users "${userIds}"`)
+
+            for (const message of history.messages) {
+                await gptscriptClient.addDatasetElement(process.env.GPTSCRIPT_WORKSPACE_ID, dataset.id, message.ts, "", await messageToString(webClient, message))
+            }
```
njhale, line 312: nit: it would be cool if `addDatasetElements` was exposed in the SDK so we could bulk add.

njhale (approved):

slack/src/tools.js
g-linville: Oh yeah I guess I forgot to add that to the Node and Python SDKs, lol. I'll do that today and then do another PR to update these.

PR #146, opened 2024-10-30
enhance: use workspaces for storing generating images

njhale:
Download generated images and store them with the workspace API

Note: The generate images tool now returns the file path in the workspace instead of an external URL.

### DIFF
--- a/images/src/generate.ts
+++ b/images/src/generate.ts
@@ -1,4 +1,7 @@
-import OpenAI from 'openai';
+import OpenAI from 'openai'
+import * as gptscript from '@gptscript-ai/gptscript'
+import axios from 'axios'
+import { createHash } from 'node:crypto'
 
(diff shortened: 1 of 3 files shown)


images/src/generate.ts
```diff
@@ -37,14 +40,39 @@ const generateImages = async (
...
+}
+
+async function download(client: gptscript.GPTScript, imageUrl: string): Promise<string> {
+  const response = await axios.get(imageUrl, {
+    responseType: 'arraybuffer'
+  })
+  const content = Buffer.from(response.data, 'binary')
+
+
+// Generate a SHA-256 hash of the imageURL to use as the filename
```
g-linville, line 67: super nit
```suggestion
  // Generate a SHA-256 hash of the imageURL to use as the filename
```

njhale: done

iwilltry42 (approved):

thedadams (approved):

PR #152, opened 2024-10-31
feat: (knowledge) add document index (location in file) to db

iwilltry42:
Adds:
- `docIndex` and `docsTotal` metadata keys (stored in VectorDB)
- `index` column in `documents` table

Fixes:
- PDF documents being re-chunked by textsplitter because chunkOverlap was not considered

### DIFF
--- a/knowledge/pkg/vectorstore/types/types.go
+++ b/knowledge/pkg/vectorstore/types/types.go
@@ -1,8 +1,58 @@
 package types
 
+import (
+	"slices"
+)
+
 type Document struct {
 	ID              string         `json:"id"`
 	Content         string         `json:"content"`
 	Metadata        map[string]any `json:"metadata"`
 	SimilarityScore float32        `json:"similarity_score"`
+}
+
+const (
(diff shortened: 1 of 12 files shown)


g-linville (approved):

knowledge/pkg/vectorstore/types/types.go
```diff
@@ -1,8 +1,58 @@
...
+)
+
+func mustInt(value any) int {
+	switch v := value.(type) {
+	case int:
+		return v
+	case float64:
+		return int(v)
+	default:
+		panic("Unsupported type")
```
thedadams, line 26: Are we sure it's safe to panic here?

thedadams (approved):

knowledge/pkg/vectorstore/types/types.go
iwilltry42: No.. this is mostly laziness at a point where I'm pretty sure we'll never hit that case 😅

PR #153, opened 2024-10-31
chore: update dataset code

g-linville:
This updates the dataset code all over the repo to ensure that:
- Each tool that can create a dataset always creates a dataset (rather than only creating one if there are more than 10 elements)
- Each tool is using a version of the SDK that includes the `addDatasetElements` function, since it is more efficient than calling `addDatasetElement` a bunch of times
- Some various other code cleanup stuff

I also added datasets to the Word `List Docs` tool, which didn't have it before.


google/sheets/readSpreadsheet.py
```diff
@@ -38,29 +39,29 @@ async def main():
...
-                print(f"Dataset created with ID {dataset.id} with {len(cell_values)} elements")
-                return
-            except Exception:
-                pass  # Ignore errors if we got any, and just print the results.
-
-        for cell_reference, value in cell_values.items():
-            if show_cell_ref:
-                print(f"{cell_reference}: {value}")
-            else:
-                print(value)
```
g-linville, line 63: Note to self: remove the corresponding tool param for this if it is no longer needed

njhale (approved):

thedadams (approved):

PR #155, opened 2024-10-31
chore: outlook calendar: make search case insensitive

g-linville:
I think this will do better.


njhale (approved):

thedadams (approved):
Makes me wonder why there isn't a `strings.ContainsFold`.

PR #156, opened 2024-10-31
enhance: slack: add user context

g-linville:
This provides additional context for the Slack tools about who the signed in user is, including their default time zone, so when they ask for things like "list all messages from today in the general channel", the LLM will know what time zone to use.

### DIFF
--- a/slack/tool.gpt
+++ b/slack/tool.gpt
@@ -174,9 +174,17 @@ Param: limit: the number of messages to return
 #!/usr/bin/env node ${GPTSCRIPT_TOOL_DIR}/index.js getDMThreadHistory
 
 ---
+Name: User Context
+Description: Get information about the logged in user and default time zone.
+Credential: ./credential
+
+#!/usr/bin/env node ${GPTSCRIPT_TOOL_DIR}/index.js userContext
+
+---
 Name: Slack Context
 Type: context
 Share Context: ../time
+Share Tools: User Context
 
 #!sys.echo
 
@@ -186,6 +194,8 @@ You have access to a set of tools to interact with a Slack workspace.
 
 Do not provide channel, thread, or message IDs in the output, as these are not helpful for the user.
 When you use the search_messages tool, you can use normal Slack search filters. If you filter by user, use the full username, which can be obtained from the list_users or search_users tools.
+
+If you are asked about messages in a particular time frame, call the User Context tool to make sure you know which time zone to use.
 
 ## End of instructions for using Slack tools
 
--- a/slack/tool.gpt
+++ b/slack/tool.gpt
@@ -183,8 +183,7 @@ Credential: ./credential
 ---
 Name: Slack Context
 Type: context
-Share Context: ../time
-Share Tools: User Context
+Share Context: ../time, User Context
 
 #!sys.echo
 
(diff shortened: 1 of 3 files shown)


slack/tool.gpt
```diff
@@ -187,6 +187,13 @@ When you use the search_messages tool, you can use normal Slack search filters.
 
 ## End of instructions for using Slack tools
 
+---
+Name: Slack User Context
+Type: context
+Credential: ./credential
+
+#!/usr/bin/env node ${GPTSCRIPT_TOOL_DIR}/index.js userContext
```
njhale, line 195: Will this get called for every tool call now? Seems like it only needs to be called once per thread in most cases.

njhale (approved):
One question, other than that LGTM

slack/tool.gpt
g-linville: I'm not sure. I tried to set this up the same way that the existing `Slack Context` tool works. The only real difference between the two is that this one runs code (and uses a credential), whereas the other one is just sys.echo.

drpebcak: Pretty sure it will be called every time - although perhaps cache will catch it?

g-linville: Is there a way I should structure this differently to avoid the excess calls?

drpebcak: I had to do something similar for hubspot in a few places.. instead of making the context be a code tool like this, I just added context tool that told it to call a specific tool to get info whenever it was doing operations that needed it. So for slack, you might have a context tool that says `Call the "Get User Context" tool to retrieve information about the currently signed in user including timezone, blah, blah. You must call this tool whenever a request is made that relates to dates/times blah blah blah...`

drpebcak: Of course it also looks like I did this exact same thing to look up user identity in the hubspot tools, so really YMMV if it causes any performance issues.

As an example though - https://github.com/otto8-ai/tools/blob/8f1d8906a3688210c49d1b63cee520a44581cd83/apis/hubspot/crm/write/tool.gpt#L18

njhale: For the `Outlook Calendar` bundle I shared the `Get Default Timezone` tool from the context and added [this line](https://github.com/otto8-ai/tools/blob/c03ab65f0d22660e733a897db4486582ddbc5974/outlook/calendar/tool.gpt#L140) to coax the LLM to get the default timezone only when necessary.

thedadams (approved):
Looks good after conflicts are fixed.

g-linville:
Re-requesting review because I made this be a callable tool instead of a context tool, to address Taylor and Nick's concerns.

Sorry for the force push; I had to rebase on main to get the dataset changes on this branch so that I can test this properly.

g-linville:
I just thought of something. Instead of what I was doing before (adding the User Context as a `Share Context` on everything), what if I added it as a `Share Context` on the main Slack Context? Then it should only get called once, right? I think that would be ideal.

cc @drpebcak let me know what you think

drpebcak:
That does sound pretty good. I imagine people will get prompted to auth with slack immediately when they open a chat that has these tools, but I don't think that's too big a deal.

g-linville:
Thanks. I just changed it to do that, and it works well. This is ready for review at last, lol.

drpebcak (approved):

PR #158, opened 2024-11-01
fix: store generated images in datasets instead of workspace files

njhale:
Use new node-gptscript method to store generated image binary data in datasets instead of as a workspace file.

### DIFF
--- a/images/package-lock.json
+++ b/images/package-lock.json
@@ -5,7 +5,7 @@
   "packages": {
     "": {
       "dependencies": {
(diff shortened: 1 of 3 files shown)


g-linville (approved):

g-linville:
Lgtm. I'll let you know when I think we're ready to merge this.

thedadams (approved):

njhale:
We decided it makes more sense to write a file to the workspace, in part so that generated images can be "round-tripped" to vision

PR #165, opened 2024-11-04
feat: add otto8 model provider

thedadams:


njhale (approved):

g-linville (approved):

drpebcak (approved):
My only concern about this is that it doesn't leave logical room for the bedrock version of the provider

PR #171, opened 2024-11-06
fix: add context to prevent notion from creating unwanted pages

njhale:
Add context to the Notion tools to confirm with the user before creating new pages. This will prevent pages from being created when the user is actually attempting to update the content of an existing page.

This change also bumps node-gptscript to include the new dataset changes.

Addresses https://github.com/otto8-ai/otto8/issues/251

### DIFF
--- a/notion/tool.gpt
+++ b/notion/tool.gpt
@@ -94,6 +94,7 @@ Type: Context
 
 You have access to a set of tools to interact with Notion.
 You cannot use the Create Page tool to update the contents of a page. It is only for creating new pages.
+Before creating a new page, confirm that the user wants to create a new page.
 
 ## End of instructions for using Notion tools
 


njhale:
Running into some dataset issues. Going to see if they're resolved with https://github.com/gptscript-ai/node-gptscript/pull/102

notion/package.json
```diff
@@ -7,7 +7,7 @@
   "author": "",
   "license": "Apache-2.0",
   "dependencies": {
-    "@gptscript-ai/gptscript": "github:gptscript-ai/node-gptscript#b30db3cf4fe87b1ee09a20752baf2263b1e5c7be",
+    "@gptscript-ai/gptscript": "github:g-linville/node-gptscript#49125f86e4a0c2af02b85373fdf647699dfcd58b",
```
drpebcak, line 10: I'd rather just wait until grant's changes are merged rather than using his fork in the primary tool repo

drpebcak (changes requested):

notion/package.json
njhale: OK -- the dataset changes have merged and I updated this dependency.

thedadams (approved):

g-linville (approved):
This LGTM, but I would prefer for it to merge after my PR to this repo, which will conflict with this one but updates all tools in a standard way.

PR #176, opened 2024-11-07
enhance: outlook mail: include default time zone

g-linville:
This is useful if the user asks some time-related question, like "list inbox messages from last week" or something.

### DIFF
--- a/outlook/mail/credential/tool.gpt
+++ b/outlook/mail/credential/tool.gpt
@@ -4,6 +4,7 @@ Share Credential: ../../../oauth2 as outlook.mail.read
         microsoft365 as integration and
         "Mail.Read
         User.Read
+        MailboxSettings.Read
         offline_access" as scope
 Type: credential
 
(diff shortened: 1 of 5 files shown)


iwilltry42 (approved):

outlook/mail/credential/tool.gpt
```diff
@@ -4,6 +4,7 @@ Share Credential: ../../../oauth2 as outlook.mail.read
         microsoft365 as integration and
         "Mail.Read
         User.Read
+        MailboxSettings.Read
```
thedadams, line 7: Just curious: does this mean that any existing credentials won't work? Or does this change of scope get recognized and the credential re-requested?

thedadams (approved):

outlook/mail/credential/tool.gpt
g-linville: Correct. Unfortunately, credentials will have to be manually deleted and recreated. GPTScript's credentials framework does not store information about OAuth scopes.

PR #193, opened 2024-11-10
chore: limit browser to getting page content

njhale:
Temporarily remove all tools other than Get Page Contents from from the Browser bundle. This is the core functionality we want to demo.


njhale:
We decided to leave these tools in since they're working. Closing out.

PR #201, opened 2024-11-12
fix: browser screenshot rendering

njhale:
Fix browser screenshot rendering by returning well-structured JSON containing screenshot `downloadUrl` fields and other info.

Bonus features:
- Add a `fullPage` parameter to the `Screenshot` tool to enable full page screenshots.
- Add a `followMode` field and some context to the Browser tools that allows users to get and display a screenshot of the browser page at the end of each successful tool call by asking the assistant to "enable follow mode"

Addresses https://github.com/otto8-ai/otto8/issues/497 for the browser tool.

### DIFF
--- a/browser/src/screenshot.ts
+++ b/browser/src/screenshot.ts
@@ -1,23 +1,38 @@
 import { type Page } from 'playwright'
 import * as gptscript from '@gptscript-ai/gptscript'
-import { getWorkspaceId } from './session.ts'
+import { getWorkspaceId, getGPTScriptEnv } from './session.ts'
 import { type IncomingHttpHeaders } from 'node:http'
 
 const client = new gptscript.GPTScript();
+const ottoServerUrl = process.env.OTTO_SERVER_URL
 
-export async function screenshot (page: Page, headers: IncomingHttpHeaders): Promise<string> {
-  const workspaceId = getWorkspaceId(headers['x-gptscript-env'])
-  let screenshotName = `screenshot-${Date.now()}_${page.url().replace(/[^a-zA-Z0-9]/g, '_')}.png`
+export async function screenshot (page: Page, headers: IncomingHttpHeaders, tabID: string, fullPage: boolean = false): Promise<string> {
   // Detect if we are running in otto8
-  if (process.env.OTTO8_THREAD_ID !== undefined) {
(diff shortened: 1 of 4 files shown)


browser/src/screenshot.ts
```diff
@@ -1,23 +1,38 @@
 import { type Page } from 'playwright'
 import * as gptscript from '@gptscript-ai/gptscript'
-import { getWorkspaceId } from './session.ts'
+import { getWorkspaceId, getGPTScriptEnv } from './session.ts'
 import { type IncomingHttpHeaders } from 'node:http'
 
 const client = new gptscript.GPTScript();
+const ottoServerUrl = process.env.OTTO_SERVER_URL
```
thedadams, line 7: The environment variable is set the same way the `OTTO_THREAD_ID` variable is. So, I would expect that "getting" it would be the same too.

njhale: Hmm, this is working locally, and I wouldn't expect it to be different when returned via the header, so is there any reason to get it from the header on every request instead of once at daemon startup?

browser/src/screenshot.ts
```diff
@@ -1,23 +1,52 @@
...
+import { getWorkspaceId, getGPTScriptEnv } from './session.ts'
 import { type IncomingHttpHeaders } from 'node:http'
+import { createHash } from 'node:crypto';
 
 const client = new gptscript.GPTScript();
+const ottoServerUrl = process.env.OTTO_SERVER_URL
 
-export async function screenshot (page: Page, headers: IncomingHttpHeaders): Promise<string> {
-  const workspaceId = getWorkspaceId(headers['x-gptscript-env'])
-  let screenshotName = `screenshot-${Date.now()}_${page.url().replace(/[^a-zA-Z0-9]/g, '_')}.png`
```
njhale, line 10: Note: I changed this because it was producing invalid `downloadURLs` for certain pages and we're exposing the page URL as part of the `ScreenshotInfo` the tool now responds with.

cloudnautique (approved):
LGTM!

thedadams (approved):

browser/src/screenshot.ts
njhale: You've convinced me. Code updated.

PR #202, opened 2024-11-12
Chore: Enforce scrape limit to 250 for now

StrongMonkey:
This PR enforce scrape limit to 250 by default now so that it will reasonably stop at 250 even when user put a bad website like cnn.com

### DIFF
--- a/knowledge/data-sources/website/main.go
+++ b/knowledge/data-sources/website/main.go
@@ -6,7 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"strings"
+	"strconv"
 
 	"github.com/gptscript-ai/go-gptscript"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


iwilltry42 (approved):

knowledge/data-sources/website/main.go
```diff
@@ -63,15 +64,15 @@ func main() {
 
 	inputData := os.Getenv("GPTSCRIPT_INPUT")
 	input := MetadataInput{}
-	for i := range input.WebsiteCrawlingConfig.URLs {
```
thedadams, line 66: This was put in to fix an issue where spaces in the URL caused errors. Has this been addressed somewhere else?

thedadams (approved):

knowledge/data-sources/website/main.go
StrongMonkey: Yeah, i addressed this in client side. It is better to address in client side because we have other checks that adds protocol automatically.

PR #212, opened 2024-11-16
feat: add embeddings support for Azure OpenAI

thedadams:


njhale (approved):

iwilltry42 (approved):

azure-openai-model-provider/main.py
```diff
@@ -72,15 +75,9 @@ async def chat_completions(request: Request):
...
-    config = await helpers.get_azure_config(data["model"])
-    if config == None:
-        raise HTTPException(status_code=400,
-                            detail="Azure config not found. Please ensure you have configured the environment variables correctly.")
-
-    client = helpers.client(
-        endpoint=config.endpoint,
-        deployment_name=config.deployment_name,
-        api_key=config.api_key
+    client = azure_client(
```
drpebcak, line 78: Looks like you aren't passing `deployment_name` in here anymore, but I'm pretty sure that's required.

azure-openai-model-provider/main.py
```diff
@@ -105,12 +102,46 @@ async def chat_completions(request: Request):
         raise HTTPException(status_code=error_code, detail=f"Error occurred: {error_message}")
 
 
+@app.post("/v1/embeddings")
+async def embeddings(request: Request):
+    data = await request.body()
+    data = json.loads(data)
+
+    client = azure_client(
```
drpebcak, line 110: Here also, doesn't this require the deployment name?

azure-openai-model-provider/tool.gpt
```diff
@@ -1,6 +1,6 @@
 Name: Azure OpenAI
 Description: Model provider for Azure OpenAI hosted models
-Metadata: envVars: OTTO8_AZURE_OPENAI_MODEL_PROVIDER_API_KEY,OTTO8_AZURE_OPENAI_MODEL_PROVIDER_ENDPOINT,OTTO8_AZURE_OPENAI_MODEL_PROVIDER_DEPLOYMENT_NAME
```
drpebcak, line 3: Here as well, you are getting rid of it. Pretty sure it's required for chat completions stuff.. is it not working for embeddings or something?

drpebcak (changes requested):

azure-openai-model-provider/tool.gpt
thedadams: The deployment name isn't required. Deployment name is Azure's replacement for model, which is still being passed. My understanding is that only one of deployment name or models needs to be set.

Rest assured that I did test these changes with both chat completions and embeddings and they worked fine.

thedadams:
I made a small change to only construct the `azure_client` once.

I also tested one more time, just to be sure, that these changes work for chat completions and providing the deployment name is not required.

drpebcak (approved):

PR #216, opened 2024-11-19
feat: add mvp bluesky tools

njhale:
Add Bluesky tools with support for basic search via unauthenticated requests.

Support for authenticated requests will require a custom OAuth provider that can handle [ATProto's authz requirements](https://atproto.com/specs/oauth#clients).

Support for datasets will be implemented as a follow up.

Depends on https://github.com/otto8-ai/otto8/pull/629

### DIFF
--- a/bluesky/src/posts.ts
+++ b/bluesky/src/posts.ts
@@ -1,0 +1,46 @@
+import { AtpAgent, AppBskyFeedSearchPosts } from '@atproto/api'
+
+export async function searchPosts (
+    agent: AtpAgent,
+    query?: string,
+    since?: string,
+    until?: string,
+    limit?: string,
+): Promise<void> {
+    let queryParams: AppBskyFeedSearchPosts.QueryParams = {
+        q: query ?? '',
+        sort: 'latest',
+    }
+
+    if (!query) {
+        throw new Error('Query is required')
+    }
+
+    if (!!limit) {
+        try {
+            queryParams.limit = parseInt(limit, 10)
+        } catch (error: unknown) {
+            throw new Error(`Invalid limit format: ${String(error)}`)
+        }
+    }
+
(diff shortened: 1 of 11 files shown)


bluesky/tool.gpt
```diff
@@ -0,0 +1,43 @@
...
+Param: query: A Lucene query to search for users with.
+Param: limit: (optional) The maximum number of users to return. Must be an integer >=1 and <=100. Defaults to 25.
+
+#!http://service.daemon.gptscript.local/tools/users/search
+
+---
+Name: service
+Metadata: index: false
+
+#!sys.daemon /usr/bin/env npm --prefix ${GPTSCRIPT_TOOL_DIR} run server
```
thedadams, line 35: Interesting that you chose to do this as a daemon. I don't think this works when we add authentication.

njhale: Oh, I had assumed the creds would get passed with the env header. I'll switch to a non-daemon

thedadams (approved):

bluesky/src/posts.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    if (limit !== undefined && limit !== '') {
+        try {
+            queryParams.limit = parseInt(limit, 10)
+        } catch (e: unknown) {
+            throw new Error(`Invalid limit format: ${String(e)}`)
+        }
+    }
+
+    if (until !== undefined && until !== '') {
```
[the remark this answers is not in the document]
njhale: Done

bluesky/tool.gpt
njhale: OK -- updated this to be a "command driven" tool

ryanhopperlowe (approved):

PR #225, opened 2024-11-21
Model provider updates

thedadams:


voyage-model-provider/main.py
```diff
@@ -0,0 +1,72 @@
+import json
+import os
+
+from voyageai import AsyncClient
+from fastapi import FastAPI, Request, HTTPException
+from fastapi.encoders import jsonable_encoder
+from fastapi.responses import JSONResponse
+
+debug = os.environ.get("GPTSCRIPT_DEBUG", "false") == "true"
+client = AsyncClient(api_key=os.environ.get("OTTO8_VOYAGE_MODEL_PROVIDER_API_KEY", "pa-P3XFfrvvOj1hzJSdUmkV_O7RVngUQ-00HpsDzgRX8sc"))
```
njhale, line 10: There’s a hardcoded key here. Was this intentional?

voyage-model-provider/main.py
```diff
@@ -0,0 +1,72 @@
...
+
+@app.get("/")
+@app.post("/")
+async def root():
+    return uri
+
+
+@app.get("/v1/models")
+async def list_models() -> JSONResponse:
+    return JSONResponse({"data": [{"id": voyage_models}]}, status_code=200)
```
njhale, line 44: Looks like this might be slightly off:
```suggestion
    return JSONResponse({"data": voyage_models}, status_code=200)
```

njhale (changes requested):

voyage-model-provider/main.py
thedadams: Good lord. What was I doing. No.

voyage-model-provider/main.py
thedadams: Good catch. Thanks.

drpebcak (approved):

PR #227, opened 2024-11-21
fix: list totals when working with datasets

drpebcak:
depends on https://github.com/gptscript-ai/datasets/pull/13


njhale (approved):
See https://github.com/gptscript-ai/datasets/pull/13#pullrequestreview-2453306935 for my thoughts on this approach.

thedadams (approved):

PR #229, opened 2024-11-22
fix: bluesky time-bound query issues 

njhale:
The `Bluesky` bundle's `Search Post` tool fails to search for posts when the `since` or `until` parameters are set due an error returned by the public bsky search APIs. This isn't well documented, but after some testing it was determined that this only occurs when making
unauthenticated requests.

To fix this issue, this change updates the Bluesky tools to make authenticated requests with a user handle and [App Password](https://bsky.app/settings/app-passwords) provided by the user via a basic auth credential tool.

This change also exposes a `tags` parameter on the `Search Post` tool to make it easier for agents to search by tag.

**Note: This PR depends on #228**

### DIFF
--- a/bluesky/credential/tool.gpt
+++ b/bluesky/credential/tool.gpt
@@ -1,0 +1,8 @@
+Name: Bluesky Basic Auth Credential
+Share Credential: ../../basic-auth as bluesky
+    with bluesky_handle as username_field and
+        BLUESKY_HANDLE as username_env and
+        bluesky_app_password as password_field and
+        BLUESKY_APP_PASSWORD as password_env and
+        "Enter your Bluesky handle and app password.\nTo generate a new app password, go to https://bsky.social/settings/app-passwords" as message
+Type: credential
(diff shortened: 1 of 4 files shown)


njhale:
Will rebase on `main` after #228 merges.

basic-auth/main.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return "", "", fmt.Errorf("Error running GPTScript prompt: %w", err)
+	}
+
+	res, err := run.Text()
+	if err != nil {
+		return "", "", fmt.Errorf("Error getting GPTScript response: %w", err)
+	}
+
+	username := gjson.Get(res, in.UsernameField).String()
```
StrongMonkey, line 84: first time seeing this gjson library? does it actually throw an error if res is not valid json?

StrongMonkey (approved):

drpebcak (approved):

PR #233, opened 2024-11-25
fix: ignore PDFs if their links lead to non-application/pdf content-types (usually redirects)

iwilltry42:
Ref https://github.com/otto8-ai/otto8/issues/150


StrongMonkey (approved):
thank you for putting up the fix! 👍

njhale (approved):

PR #236, opened 2024-11-26
Fix: Properly wrap table names and values

cjellick:
In otto8, table names and values have a / in them and sqlite didn't like
that.

Signed-off-by: Craig Jellick <craig@acorn.io>


iwilltry42 (commented):
Is there a reason to use `[%s]` over `"%s"` (which could be `%q`)?
As per SQLite docs, `[ ]` is non-standard SQL and only there for compatibility reasons.
Doesn't harm I guess, just asking :thinking:

thedadams (approved):

PR #238, opened 2024-11-26
Fix: put parent page into its own folder

StrongMonkey:
https://github.com/otto8-ai/otto8/issues/667

For any parent page that could have children, place it into the folder itself. This is to avoid parent page ending in a separate node in a tree.

### DIFF
--- a/knowledge/data-sources/notion/src/index.ts
+++ b/knowledge/data-sources/notion/src/index.ts
@@ -73,6 +73,7 @@ async function getAllPagesIteratively(
         {
             page: PageObjectResponse;
             path: string;
+            hasChildren: boolean;
         }
     >();
     const stack: (PageObjectResponse | null)[] = [];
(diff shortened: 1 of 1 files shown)


knowledge/data-sources/notion/src/index.ts
```diff
@@ -109,6 +110,7 @@ async function getAllPagesIteratively(
         pages.set(currentPage.id, {
             page: currentPage,
             path: "",
+            hasChildren: false,
```
g-linville, line 113: Is this call to `pages.set` needed anymore? Since we call it again after computing `hasChildren` in the do-while loop below, I don't think this is needed here anymore.

StrongMonkey: Oh I didn't do this in the do while because this is more like an override, and also it will also need to handle the case of failure if get children call failed. So it is probably better to handle it beforehand and override it in do while loop.

g-linville (approved):

PR #246, opened 2024-11-26
enhance: migrate sqlite-vec code to gorm and use batched transactions

iwilltry42:
There's likely room for improvement, but I was too tired to continue.

By batching the write queries, the concurrent setup in otto8 should get faster and we have less blocking by locks.

Ref https://github.com/otto8-ai/otto8/issues/684


knowledge/pkg/vectorstore/sqlite-vec/sqlite-vec.go
```diff
@@ -43,23 +47,18 @@ func New(ctx context.Context, dsn string, embeddingFunc cg.EmbeddingFunc) (*Vect
...
+		return nil, fmt.Errorf("failed to scan results: %w", err)
 	}
+	slog.Debug("sqlite-vec info", "sqlite_version", sqliteVersion, "vec_version", vecVersion)
 
 	return store, store.prepareTables(ctx)
 }
 
 func (v *VectorStore) Close() error {
-	return v.db.Close()
+	return nil
```
thedadams, line 61: Not sure how important this is, but it is possible to get the underlying database connection from gorm and close it.

thedadams (approved):

knowledge/pkg/vectorstore/sqlite-vec/sqlite-vec.go
iwilltry42: I remembered a comment about it not being recommended: https://github.com/go-gorm/gorm/issues/3145#issuecomment-658230163
.. however, since we run knowledge as one-off commands, this is certainly different for us.
I figured I never called any Close() method anywhere, so I now added it all throughout the code :+1:

StrongMonkey (approved):

PR #248, opened 2024-11-27
feat: add a semi-consistent models api to each provider

thedadams:
This change allows Otto8 to call the models APIs to get the available models for each provider.

One thing to note here: this adds some additional complexity to the Azure OpenAI provider with additional configuration.


azure-openai-model-provider/tool.gpt
```diff
@@ -1,6 +1,6 @@
 Name: Azure OpenAI
 Description: Model provider for Azure OpenAI hosted models
-Metadata: envVars: OTTO8_AZURE_OPENAI_MODEL_PROVIDER_API_KEY,OTTO8_AZURE_OPENAI_MODEL_PROVIDER_ENDPOINT
+Metadata: envVars: OTTO8_AZURE_OPENAI_MODEL_PROVIDER_API_KEY,OTTO8_AZURE_OPENAI_MODEL_PROVIDER_ENDPOINT,OTTO8_AZURE_OPENAI_MODEL_PROVIDER_CLIENT_ID,OTTO8_AZURE_OPENAI_MODEL_PROVIDER_CLIENT_SECRET,OTTO8_AZURE_OPENAI_MODEL_PROVIDER_TENANT_ID,OTTO8_AZURE_OPENAI_MODEL_PROVIDER_SUBSCRIPTION_ID,OTTO8_AZURE_OPENAI_MODEL_PROVIDER_RESOURCE_ID
```
drpebcak, line 3: If we are making these mandatory, should we just go ahead and require the user to configure this for the OpenAI client's usage too, and drop support for api key? That would solve the issue of requiring one of the two methods.

thedadams: Do you mean require that the service principal has permission to do chat completion and whatnot?

g-linville (approved):

azure-openai-model-provider/tool.gpt
thedadams: Yeah, I didn't think of that. And there is a built-in role for that, if I remember.

If you're around today, can you give my service principal that role and I will test this out when I'm back.

drpebcak: I gave it the role, but I also did some testing - it works. So we just have to decide if this is the route we want to go down or not.

drpebcak: https://github.com/thedadams/otto8-tools/pull/1

thedadams: @drpebcak I added your commit to this change, too. Thanks for doing that.

PR #249, opened 2024-11-29
fix: use temporary userprofiles for libreoffice/soffice during conversion so it doesn't break when run concurrently

iwilltry42:
By default, multiple concurrent calls to `soffice` will use the same user profile, thus try to hand off a conversion task to an already running instance - which fails in our scenario, since we're running adhoc conversions in headless mode.

Option A: Run libreoffice "server mode" and send conversion tasks there, e.g. via unoconv or similar
Option B: Create temporary profile directories for each conversion task so they don't conflict <- We're doing this here

Ref https://github.com/otto8-ai/otto8/issues/630

### DIFF
--- a/knowledge/pkg/datastore/documentloader/converter/soffice.go
+++ b/knowledge/pkg/datastore/documentloader/converter/soffice.go
@@ -53,20 +53,35 @@ func (c *SofficeConverter) Convert(ctx context.Context, reader io.Reader, source
 	}
 	_ = tempfile.Close()
 
+	logger := log.FromCtx(ctx)
+
+	profileDir, err := os.MkdirTemp(os.TempDir(), "libreoffice-profile-*")
+	if err != nil {
+		logger.Error("Failed to create soffice profile directory", "path", profileDir, "error", err)
+		return nil, fmt.Errorf("failed to create soffice profile directory: %w", err)
+	}
+	defer os.RemoveAll(profileDir)
+
 	// Convert the file using soffice
-	cmd := exec.Command("soffice", "--headless", "--convert-to", outputFormat, "--outdir", os.TempDir(), p)
+	cmd := exec.Command(
+		"soffice",
+		"--headless",
+		fmt.Sprintf("-env:UserInstallation=file://%s", profileDir),
+		"--convert-to", outputFormat,
+		"--outdir", os.TempDir(),
+		p,
+	)
 
 	// capture stdout and stderr in a buffer
 	var outb, errb strings.Builder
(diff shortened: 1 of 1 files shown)


knowledge/pkg/datastore/documentloader/converter/soffice.go
```diff
@@ -53,20 +53,35 @@ func (c *SofficeConverter) Convert(ctx context.Context, reader io.Reader, source
 	}
 	_ = tempfile.Close()
 
+	logger := log.FromCtx(ctx)
+
+	profileDir, err := os.MkdirTemp(os.TempDir(), "libreoffice-profile-*")
```
thedadams, line 58: Does it make sense to clean up any of these directories/files?

Also, it looks like this is just an empty file? Is that intended?

thedadams (approved):

knowledge/pkg/datastore/documentloader/converter/soffice.go
iwilltry42: Cleanup is deferred: https://github.com/otto8-ai/tools/pull/249/files/4bd30a02c44e367c1321019c483c41f8f5ebe993#diff-a3f56052448642b16aaeabd7d8a58f4e7c4251780db6ce377f91dbd915d48eaaR63

> Also, it looks like this is just an empty file? Is that intended?

It's just an empty dir, that's correct. `soffice` initializes it on start.
In fact, the dir doesn't even need to exist before - it just must not be used by another instance of libreoffice at the same time.
There could be smarter solutions like using a server or just some "connection pooling" with a fixed set of profiles, but I'd like to defer that extra engineering overhead for now.

knowledge/pkg/datastore/documentloader/converter/soffice.go
```diff
@@ -53,20 +53,35 @@ func (c *SofficeConverter) Convert(ctx context.Context, reader io.Reader, source
...
+		return nil, fmt.Errorf("failed to create soffice profile directory: %w", err)
+	}
+	defer os.RemoveAll(profileDir)
+
 	// Convert the file using soffice
-	cmd := exec.Command("soffice", "--headless", "--convert-to", outputFormat, "--outdir", os.TempDir(), p)
+	cmd := exec.Command(
+		"soffice",
+		"--headless",
+		fmt.Sprintf("-env:UserInstallation=file://%s", profileDir),
```
g-linville, line 69: Just wanted to leave a comment here to make sure that the single dash (`-env`) was intentional.

g-linville (approved):

PR #250, opened 2024-12-01
change: bring back website cleaning, but act on a copy of the DOM as to not remove links (e.g. from nav) before we checked them out.

iwilltry42:
This reverts commit 6ceec6f79cce871fae1fecc0c3b720502eba2b19.

Part of enhancement addressing https://github.com/otto8-ai/otto8/issues/386

### DIFF
--- a/knowledge/data-sources/website/clean.go
+++ b/knowledge/data-sources/website/clean.go
@@ -1,0 +1,23 @@
+package main
+
+import (
+	"github.com/PuerkitoBio/goquery"
+)
+
+var excludeNonMainTags = []string{
+	"script, style, noscript, meta, head",
+	"header", "footer", "nav", "aside", ".header", ".top", ".navbar", "#header",
+	".footer", ".bottom", "#footer", ".sidebar", ".side", ".aside", "#sidebar",
+	".modal", ".popup", "#modal", ".overlay", ".ad", ".ads", ".advert", "#ad",
+	".lang-selector", ".language", "#language-selector", ".social", ".social-media",
+	".social-links", "#social", ".menu", ".navigation", "#nav", ".breadcrumbs",
+	"#breadcrumbs", "#search-form", ".search", "#search", ".share", "#share",
+	".widget", "#widget", ".cookie", "#cookie",
+}
+
+func RemoveNonMainContent(doc *goquery.Selection) *goquery.Selection {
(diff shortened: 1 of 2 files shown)


knowledge/data-sources/website/clean.go
```diff
@@ -0,0 +1,23 @@
...
+var excludeNonMainTags = []string{
+	"script, style, noscript, meta, head",
+	"header", "footer", "nav", "aside", ".header", ".top", ".navbar", "#header",
+	".footer", ".bottom", "#footer", ".sidebar", ".side", ".aside", "#sidebar",
+	".modal", ".popup", "#modal", ".overlay", ".ad", ".ads", ".advert", "#ad",
+	".lang-selector", ".language", "#language-selector", ".social", ".social-media",
+	".social-links", "#social", ".menu", ".navigation", "#nav", ".breadcrumbs",
+	"#breadcrumbs", "#search-form", ".search", "#search", ".share", "#share",
+	".widget", "#widget", ".cookie", "#cookie",
+}
```
thedadams, line 16: Would it be easier to maintain a list or set of tags we do want?

thedadams (approved):

knowledge/data-sources/website/clean.go
iwilltry42: No idea tbh.
IIUC, we [only extract the `body` anyway](https://github.com/otto8-ai/tools/pull/250/files#diff-f548d0250be14e804c5f82dc2bd0d44bc0ba1ebc38595608d67badb4a6296931R52), meaning that a bunch of those tags are unlikely to appear within.
However, I fear like maintaining an "allowlist" would be more prone to drop things which might be actual content but were just using strange or unknown tags.
I wouldn't fight over changing this though :)

iwilltry42: FWIW, we still send this to an LLM for further cleaning (after the markdown conversion)

thedadams: I guess it depends on what is more valuable. If getting potentially extra data is the desired "failure" state, then this is the way to go. If we want to ensure we are only getting certain tags, then the allow list would be a better choice.

iwilltry42:
Closing this as we're going to rewrite the website-cleaning tool to be code-based and handle the retries itself.

PR #259, opened 2024-12-04
Fix: refactored onedrive to download file when syncing metadata

StrongMonkey:
https://github.com/otto8-ai/otto8/issues/538

Refactored the code a bit so that when fetch children files from folder, start syncing immediately so that user will see file showing up in UI.

### DIFF
--- a/knowledge/data-sources/onedrive/main.go
+++ b/knowledge/data-sources/onedrive/main.go
@@ -156,10 +156,7 @@ func main() {
 }
 
 func sync(ctx context.Context, logErr *logrus.Logger, input MetadataInput, output *MetadataOutput, client *msgraphsdk.GraphServiceClient, gptscript *gptscript.GPTScript) error {
-	items := make(map[string]struct {
-		Item models.DriveItemable
-		Root string
-	})
+	items := map[string]models.DriveItemable{}
 	for _, link := range input.OneDriveConfig.SharedLinks {
 		requestParameters := &shares.ItemDriveItemRequestBuilderGetQueryParameters{
 			Expand: []string{"children"},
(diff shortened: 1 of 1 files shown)


knowledge/data-sources/onedrive/main.go
```diff
@@ -177,26 +174,30 @@ func sync(ctx context.Context, logErr *logrus.Logger, input MetadataInput, outpu
...
+			items[*child.GetId()] = child
+		}
+	}
+
+	for id := range output.Files {
+		found := false
+		if _, ok := items[id]; ok {
+			found = true
+		}
+		if !found {
```
thedadams, line 191:
```suggestion
		if _, ok := items[id]; !ok {
```

thedadams (approved):

knowledge/data-sources/onedrive/main.go
```diff
@@ -209,8 +210,15 @@ func writeMetadata(ctx context.Context, output *MetadataOutput, gptscript *gptsc
 	return gptscript.WriteFileInWorkspace(ctx, ".metadata.json", data)
 }
 
-func getChildrenFileForItem(ctx context.Context, client *msgraphsdk.GraphServiceClient, item models.DriveItemable) ([]models.DriveItemable, error) {
+func syncChildrenFileForItem(ctx context.Context, client *msgraphsdk.GraphServiceClient, gptscriptClient *gptscript.GPTScript, item models.DriveItemable, output *MetadataOutput, root string, logErr *logrus.Logger) ([]models.DriveItemable, error) {
 	if item.GetFile() != nil {
+		// We only sync item that is less than 50 MB, as most of the bigger files won't be supported from knowledge
```
iwilltry42, line 215: Should we rather add a check if the filetype is likely to be supported? Does OneDrive give a mimetype for files?

iwilltry42 (approved):

PR #264, opened 2024-12-05
fix: convert excel serials to dates

drpebcak:


excel/main.go
```diff
@@ -33,6 +35,17 @@ func main() {
 		err = commands.AddWorksheetRow(context.Background(), os.Getenv("WORKBOOK_ID"), os.Getenv("WORKSHEET_ID"), os.Getenv("CONTENTS"))
 	case "createWorksheet":
 		err = commands.CreateWorksheet(context.Background(), os.Getenv("WORKBOOK_ID"), os.Getenv("NAME"))
+	case "getDate":
+		serialStrings := strings.Split(os.Getenv("SERIALS"), "|")
+		serials := make([]int, len(serialStrings))
+		for i, v := range serialStrings {
+			val, err := strconv.Atoi(strings.TrimSpace(v))
+			if err != nil {
+				break
```
g-linville, line 44: I feel like we should print an error here, and then either continue with the loop or just exit(1).

drpebcak: Hm.. yeah i dont think I thought this through all the way because it doesn't break all the way out of the switch. I've just pushed a change that should address this.

g-linville (approved):

excel/pkg/commands/getDate.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+func GetDate(serials []int) {
+	startDate := time.Date(1900, time.January, 1, 0, 0, 0, 0, time.UTC)
+	for _, v := range serials {
+		date := startDate.AddDate(0, 0, v-2).Format(time.DateOnly)
```
thedadams, line 11: This is a little confusing. I would appreciate a comment here.

thedadams (approved):

PR #265, opened 2024-12-05
enhance: update well-known sqlite db workspace file path

njhale:
- Change the workspace directory of the sqlite workspace DB file to
`/databases/`
- Refactor workspace save logic to avoid unnecessary hash and file read
  operations


sqlite/main.go
```diff
@@ -75,7 +73,11 @@ func main() {
 	var result string
 	switch command {
 	case "exec":
+		// Calculate hash of the initial data
```
g-linville, line 76: looks like this comment shouldn't be here

njhale: Comment terminated!

PR #266, opened 2024-12-06
fix: CreatePage tool fix -- split content if it exceeds length limit 2000

tybalex:
This PR aims to fix issue https://github.com/otto8-ai/otto8/issues/627.

*What is fixed*:
the `CreatePage` tool now will check the length of input content, and if it's greater than the limit(2000), the tool will slice the content into smaller chunks as multiple rich_texts of the the children block of the page.

Screenshot 2024-12-06 at 8 23 36 AM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 **/venv
 **/bin
 **/node_modules
+*__pycache__
--- a/notion/src/pages.js
+++ b/notion/src/pages.js
@@ -51,12 +51,9 @@ function sliceAndMergeContents(contents, maxLength = 2000) {
 export async function createPage(client, name, contents, parentPageId) {
     const contentChunks = sliceAndMergeContents(contents);
 
-    const children = contentChunks.map(chunk => ({
-        object: "block",
-        type: "paragraph",
-        paragraph: {
-            rich_text: [{ type: "text", text: { content: chunk } }],
-        },
+    const richTexts = contentChunks.map(chunk => ({
+        type: "text",
+        text: { content: chunk.trim() }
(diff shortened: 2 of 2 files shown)


cjellick:
Does this end up looking like separate blocks? Like, if the post is one long paragraph that goes over 2k characters, will that paragraph get split into separate blocks?

tybalex:
> Does this end up looking like separate blocks? Like, if the post is one long paragraph that goes over 2k characters, will that paragraph get split into separate blocks?

Yes, a long paragraph without `\n` will be spliced by `.`, in this case this paragraph gets split into separate blocks.

tybalex:
Just made an update such that the function will create multiple `rich_text` instead of creating multiple blocks. So the user won't even tell the text got split.
@cjellick Tested this approach with a long paragraph and looks like it works well. The text does not get split into 2 paragraphs in notion page.
Screenshot 2024-12-06 at 10 33 47 AM

g-linville (approved):
If you're willing, can you remove the semicolons from the new code? It looks like the other code for this tool does not use semicolons. I wish JavaScript either explicitly required or forbade them, instead of making them optional, lol.

cjellick:
@g-linville you should go ahead and request chanes rather than approve this. i think its wort being explicit and intentional about enforcing something like that

g-linville (changes requested):
Requesting changes for semicolons to be removed

tybalex:
> Requesting changes for semicolons to be removed

Removed all semicolons as requested.

g-linville (approved):

PR #272, opened 2024-12-06
refactor: rename sqlite bundle

njhale:
Remove user-facing implementation details from the `SQLite` bundle by
renaming it to `Database`.

References to "SQLite" are kept in tool descriptions and context because
it informs the LLM that their arguments must be valid SQLite.


database/go.mod
```diff
@@ -1,4 +1,4 @@
-module sqlite
+module otto8-ai/database
```
drpebcak, line 1: Why does the module name reference otto8 now?

njhale: To avoid conflicts with https://pkg.go.dev/database. Seeing `database/sql` next to `database/pkg/cmd` in the import statement felt unholy. Didn't check if it would actually cause resolution issues since there's no overlap in the package paths after `database`, but it seemed confusing enough to probably be an anti-pattern at least.

I can change it back or come up with a new module name if you'd like, but it should probably stay relative in case we ever share code from this with other tools in the repo . IUUC, it's already the precedent since it makes it easier to work with in mono-repos (i.e. I don't need to push changes to main before I can use them in dependents)

cjellick (approved):

g-linville (approved):

database/go.mod
njhale: going to merge this and we can follow up if we want to change it

PR #274, opened 2024-12-08
enhance: mark sensitive model provider parameters

njhale:
Mark sensitive model provider parameters with the `!` prefix to hide their values in the Otto8 model provider config form.

Depends on https://github.com/otto8-ai/otto8/pull/824


cjellick:
Don't merge this. I don't see an upside to the backend doing this

thedadams:
I agree with @cjellick. Doesn't seem worth the headache of trying explain why the `!` exists.

PR #275, opened 2024-12-08
fix: support copying generated image files

njhale:
Add the `workspace_copy` tool to the `Workspace Files` bundle in order to support copying workspace files without exposing their content to the LLM. This allows agents to copy large and/or non-UTF8 files and enables save generated images to files with more descriptive names.

Addresses https://github.com/otto8-ai/otto8/issues/514

**Note:** ~This change has me wondering if the `workspace_read` tool should output markdown with the URL format we use for displaying images in the `Images` bundle; e.g. -> `generated_image_4663217f.png` -> ``. The idea is for the UI to show the image on read instead of failing.~

**Note:** I’m feeling like this is the wrong approach since having the LLM read/write non-utf8 files directly isn’t tenable. I’m coming around to the idea that hiding these files and providing first-class naming/renaming support in the Images bundle is a more consistent solution.

### DIFF
--- a/workspace-files/main.go
+++ b/workspace-files/main.go
@@ -23,8 +23,8 @@ var (
 func main() {
 	if len(os.Args) == 1 {
 		fmt.Printf(`
-Subcommands: read, write
-env: FILENAME, CONTENT, GPTSCRIPT_WORKSPACE_DIR
+Subcommands: read, write, copy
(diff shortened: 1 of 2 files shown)


thedadams (approved):

g-linville (approved):

njhale:
Talked this over with Donnie and we came to the conclusion that this makes sense both as an independent feature _and_ as a solution for https://github.com/otto8-ai/otto8/issues/514, merging.

PR #278, opened 2024-12-12
fix: support the claude/anthropic provider's vision response

iwilltry42:
depends on https://github.com/gptscript-ai/claude3-provider-common/pull/5

Ref https://github.com/otto8-ai/otto8/issues/813

### DIFF
--- a/anthropic-model-provider/requirements.txt
+++ b/anthropic-model-provider/requirements.txt
@@ -1,4 +1,4 @@
 fastapi
 uvicorn[standard]
 anthropic==0.40.0
-claude3_provider_common @ git+https://github.com/iwilltry42/claude3-provider-common@30487c6fd56f66e9d7f2f1159d99478b409c6a40
+claude3_provider_common @ git+https://github.com/gptscript-ai/claude3-provider-common@2682b94fce767c810254075253ffc6e6900654fc
--- a/anthropic-model-provider/requirements.txt
+++ b/anthropic-model-provider/requirements.txt
@@ -1,4 +1,4 @@
 fastapi
 uvicorn[standard]
 anthropic==0.40.0
-claude3_provider_common @ git+https://github.com/gptscript-ai/claude3-provider-common@332d95d729e696d2f12d0d4edd7a870623764bf2
+claude3_provider_common @ git+https://github.com/iwilltry42/claude3-provider-common@30487c6fd56f66e9d7f2f1159d99478b409c6a40
--- a/knowledge/pkg/datastore/documentloader/ocr/openai/openai.go
+++ b/knowledge/pkg/datastore/documentloader/ocr/openai/openai.go
@@ -10,6 +10,8 @@ import (
 	"image/png"
 	"io"
 	"net/http"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/acorn-io/z"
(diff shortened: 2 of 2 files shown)


anthropic-model-provider/requirements.txt
```diff
@@ -1,4 +1,4 @@
 fastapi
 uvicorn[standard]
 anthropic==0.40.0
-claude3_provider_common @ git+https://github.com/gptscript-ai/claude3-provider-common@332d95d729e696d2f12d0d4edd7a870623764bf2
```
iwilltry42, line 4: TODO: change once https://github.com/gptscript-ai/claude3-provider-common/pull/5 got merged

knowledge/pkg/datastore/documentloader/ocr/openai/openai.go
```diff
@@ -61,7 +63,8 @@ type RespMessage struct {
 
 type Choice struct {
 	FinishReason string      `json:"finish_reason"`
-	Message      RespMessage `json:"message"`
+	Message      RespMessage `json:"message"` // e.g. OpenAI provider resp
+	Delta        RespMessage `json:"delta"`   // e.g. Anthropic/Claude provider StreamResponse
```
StrongMonkey, line 67: Should the logic be a seperate file like anthropic.go?

iwilltry42: This is going through the LLM Proxy anyway, so even OpenAI doesn't fit here anymore.
Also this theoretically applies for any model-provider we have or write that uses a stream response

thedadams: The point of the model providers is to make them such that the client doesn't need to know it's not OpenAI. If there is a difference between what the model provider produces and what you expect here, then the model provider should change.

iwilltry42: This will change completely with https://github.com/otto8-ai/otto8/issues/831 and https://github.com/otto8-ai/otto8/issues/832 - so I'd just leave it like this for now :+1:

iwilltry42: Follow-Up issue: https://github.com/obot-platform/obot/issues/831

knowledge/pkg/datastore/documentloader/ocr/openai/openai.go
```diff
@@ -290,10 +293,34 @@ func (o *OpenAIOCR) SendImageToOpenAI(ctx context.Context, base64Image string) (
 		return "", fmt.Errorf("OpenAI OCR error sending request(s): %w", err)
 	}
 
+	err = os.WriteFile("/tmp/openai_response.json", body, 0777)
+	if err != nil {
+		return "", fmt.Errorf("error writing openai response to file: %w", err)
+	}
```
g-linville, line 299: Was this for debugging? If so, it should be removed.

njhale (approved):
I think this should be good once [Grant's comment](https://github.com/obot-platform/tools/pull/278/files#r1888996257) is addressed.

knowledge/pkg/datastore/documentloader/ocr/openai/openai.go
iwilltry42: :sweat_smile: no idea how that could slip through

PR #279, opened 2024-12-13
fix google search tool: properly encode search query parameters

sanjay920:
Is a fix for https://github.com/otto8-ai/otto8/issues/856

- Add encodeURIComponent for Google search URL
- Improve URL safety for special characters and Unicode
- Prevent malformed search requests
- Change gptscript tool description to get llm to re-call the tool if it believes it needs more search results
- Change gptscript maxResults minimum value to 2 - openai gpt4o was setting this to 1 in many runs and this was preventing the run from finding the correct answer

### DIFF
--- a/google/search/tool.gpt
+++ b/google/search/tool.gpt
@@ -6,12 +6,12 @@ Share Tools: Search
 
 ---
 Name: Search
-Description: Search Google with a given query and return relevant information from the search results
+Description: Search Google with a given query and return relevant information from the search results. Search with more maxResults if you need more information.
 JSON Response: true
 Share Context: ../../time
 Tools: service
-Args: maxResults: The maximum number of search results to gather relevant information from (optional, default 3)
 Args: query: A question, statement, or topic to search with (required)
+Args: maxResults: The maximum number of search results to gather relevant information from (optional, default 3, minimum 2)
 
 #!http://service.daemon.gptscript.local/search
 
--- a/google/search/src/search.ts
+++ b/google/search/src/search.ts
@@ -21,6 +21,10 @@ export async function search (
   if (query === '') {
     throw new Error('No query provided')
   }
+
+  const encodedQuery = encodeURIComponent(query)
+  const searchUrl = `https://www.google.com/search?q=${encodedQuery}&udm=14`
+  
   const foundURLs = new Set<string>()
   const results: Array<Promise<SearchResult | null>> = []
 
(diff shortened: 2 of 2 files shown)


google/search/tool.gpt
```diff
@@ -6,12 +6,12 @@ Share Tools: Search
...
 ---
 Name: Search
-Description: Search Google with a given query and return relevant information from the search results
+Description: Search Google with a given query and return relevant information from the search results. Search with more maxResults if you need more information.
 JSON Response: true
 Share Context: ../../time
 Tools: service
-Args: maxResults: The maximum number of search results to gather relevant information from (optional, default 3)
 Args: query: A question, statement, or topic to search with (required)
+Args: maxResults: The maximum number of search results to gather relevant information from (optional, default 3, minimum 2)
```
njhale, line 14: The LLM isn't always great at respecting tool parameter descriptions. I'd also add some validation on the code side and return an error describing the lower (and potentially upper) bounds here.

njhale (changes requested):

google/search/tool.gpt
sanjay920: I agree with your point about validation, but I think it’s more appropriate to address this on the `gptscript` tool parameter side rather than in this PR. Specifically, we should introduce support for the `required` field for function definitions and `minimum`/`maximum` fields for numeric properties like this. For example:

```json
{
  "name": "get_n_day_weather_forecast",
  "description": "Get an N-day weather forecast",
  "strict": false,
  "parameters": {
    "type": "object",
    "properties": {
      "location": {
        "type": "string",
        "description": "The city and state, e.g., San Francisco, CA"
      },
      "format": {
        "type": "string",
        "enum": ["celsius", "fahrenheit"],
        "description": "The temperature unit to use. Infer this from the user's location."
      },
      "num_days": {
        "type": "integer",
        "description": "The number of days to forecast",
        "minimum": 3,
        "maximum": 10
      }
    },
    "required": ["location", "format", "num_days"]
  }
}
```

This kind of structure ensures validation happens at the tool definition level, keeping function parameters clean and easy to enforce boundaries. I’d be happy to explore this further in a separate PR or discussion thread!

njhale: approving, we can follow up with code-level validation if it becomes an issue.

thedadams (approved):

PR #290, opened 2024-12-18
chore: initial commit of new salesforce tools

drpebcak:
relies on changes in https://github.com/obot-platform/obot/pull/960 for salesforce integration to work

### DIFF
--- a/excel/credential/tool.gpt
+++ b/excel/credential/tool.gpt
@@ -1,6 +1,6 @@
 Name: Microsoft Excel OAuth Credential
 Share Credential: ../../oauth2 as excel.write
-    with GPTSCRIPT_MICROSOFT_EXCEL_TOKEN as env and
+    with GPTSCRIPT_MICROSOFT_EXCEL_TOKEN as token and
         microsoft365 as integration and
         "Files.Read
         Files.ReadWrite
--- a/salesforce/accounts/createAccount.py
+++ b/salesforce/accounts/createAccount.py
@@ -1,0 +1,35 @@
+import json
+import os
+import sys
+from pathlib import Path
+
+sys.path.append(str(Path(__file__).resolve().parent.parent))
+
+from helpers.auth import client
+
+
+def main():
+    try:
+        account = os.getenv("ACCOUNT")
+        if account is None:
+            print("No account provided")
+            exit(1)
+        account = json.loads(account)
+    except json.JSONDecodeError:
+        print("Invalid JSON provided")
+        exit(1)
+    except Exception as e:
+        print(f"An error occurred: {e}")
+        exit(1)
+
+    try:
+        sf = client()
+        account = sf.Account.create(account)
+        print(f"Account created successfully with Id: {account['id']}")
+    except Exception as e:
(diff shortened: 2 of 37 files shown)


salesforce/accounts/createAccount.py
```diff
@@ -0,0 +1,35 @@
+import json
+import os
+import sys
+from pathlib import Path
+
+sys.path.append(str(Path(__file__).resolve().parent.parent))
```
thedadams, line 6: This is in all the files. Does a comment make sense here? I don't know why this is needed or if this is standard practice.

thedadams (approved):

njhale (approved):

excel/credential/tool.gpt
```diff
@@ -1,6 +1,6 @@
 Name: Microsoft Excel OAuth Credential
 Share Credential: ../../oauth2 as excel.write
-    with GPTSCRIPT_MICROSOFT_EXCEL_TOKEN as env and
+    with GPTSCRIPT_MICROSOFT_EXCEL_TOKEN as token and
```
StrongMonkey, line 3: Oh, what was the context to change from env to token?

salesforce/query.py
```diff
@@ -0,0 +1,34 @@
...
+    if res.get("totalSize") > 0:
+        print_res(res)
+        next_url = res.get("nextRecordsUrl")
+        while next_url:
+            res = sf.query_more(next_url, True)
+            next_url = res.get("nextRecordsUrl")
+            print_res(res)
+    else:
+        print("No records found")
+        exit(1)
```
g-linville, line 25: nit: I don't think this makes a difference to GPTScript, but this seems like an exit(0) situation to me

```suggestion
        exit(0)
```

g-linville (approved):

salesforce/accounts/createAccount.py
drpebcak: honestly im not sure this is the right way to do this, but python is scary and confusing. it's essentially here because i organized the files into different folders but we are calling the scripts individually, not as a whole python package - so it doesnt know how to find and import the other files without this.

excel/credential/tool.gpt
drpebcak: essentially just that `env` isn't quite accurate anymore since it refers to the value used for the oauth token env var specifically, where now we support a map of extra env vars that can be added to this credential as well.

salesforce/tool.gpt
```diff
@@ -0,0 +1,232 @@
...
+Param: case_id: A string containing the Salesforce Id of the case to delete
+
+#!/usr/bin/env python3 ${GPTSCRIPT_TOOL_DIR}/cases/deleteCase.py
+
+---
+Name: Query
+Description: Query Salesforce using SOQL
+Share Context: Salesforce Context
+Credential: ./credential
+Param: query: The SOQL query to execute
```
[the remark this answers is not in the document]
drpebcak: in my experience it seems to know it pretty well, but i think we would need someone more experienced with salesforce to really put it through its paces.

PR #292, opened 2024-12-19
feat: add vLLM model provider

sanjay920:
depends on https://github.com/obot-platform/obot/pull/968


vllm-model-provider/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/sanjay920/tools/vllm-model-provider
+
+go 1.23.4
+
+require github.com/gptscript-ai/chat-completion-client v0.0.0-20241216203633-5c0178fb89ed // indirect
```
g-linville, line 5: Can you `go mod tidy` this? I don't think it should be marked as indirect.

vllm-model-provider/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/sanjay920/tools/vllm-model-provider
```
g-linville, line 1:
```suggestion
module github.com/obot-platform/tools/vllm-model-provider
```

vllm-model-provider/server/server.go
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    resp.Body = io.NopCloser(bytes.NewReader(b))
+    resp.Header.Set("Content-Length", fmt.Sprintf("%d", len(b)))
+    return nil
+}
+
+func (s *server) proxy(req *http.Request) {
+    req.URL.Host = s.endpoint
+    req.URL.Scheme = "https"
```
g-linville, line 120: I noticed that we are forcing this to be HTTPS here (and that InsecureSkipVerify gets set on the TLS config for the reverse proxy). I'm curious to know why this is. Since vLLM is something you run yourself, I could see people running it on a local network over just HTTP, or running it on a larger scale and using HTTPS (and wanting to keep verification on).

vllm-model-provider/server/server.go
```diff
@@ -0,0 +1,124 @@
...
+    req.Host = req.URL.Host
```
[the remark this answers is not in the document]
thedadams: Just realized that my suggestion won't work. It should be:
```go
    u := * req.endpoint
    u.Path = path.Join(u.Path, req.URL.Path)
    req.URL = &u
    req.Host = req.URL.Host
```

vllm-model-provider/server/server.go
sanjay920: fair point - ill default to HTTP for localhost/127.0.0.1 and HTTPS for all other hosts
fwiw i dont think users will run obot on the same machine with their GPUs

vllm-model-provider/server/server.go
```diff
@@ -0,0 +1,125 @@
...
+	"path"
+
+	openai "github.com/gptscript-ai/chat-completion-client"
+)
+
+func Run(apiKey, endpointStr, port string) error {
+	if endpointStr == "" {
+		return fmt.Errorf("endpoint is required")
+	}
+
```
thedadams, line 22: nit: already checked this, so can probably remove from here. Plus, it will be caugh below when you try to parse the URL.

vllm-model-provider/server/server.go
```diff
@@ -0,0 +1,125 @@
...
+	// Parse the endpoint URL
+	endpoint, err := url.Parse(endpointStr)
+	if err != nil {
+		return fmt.Errorf("invalid endpoint URL: %w", err)
```
thedadams, line 26:
```suggestion
		return fmt.Errorf("invalid endpoint URL %q: %w", endpointStr, err)
```

thedadams (approved):

g-linville (approved):

PR #300, opened 2024-12-23
chore: add context to permissions error for github

drpebcak:


github/tool.gpt
```diff
@@ -193,10 +209,24 @@ Tools: github.com/gptscript-ai/datasets/filter
...
 
 #!/usr/bin/env node ${GPTSCRIPT_TOOL_DIR}/index.js listPRsForReview
 
+---
+Name: GitHub Context
+Type: context
+
+#!sys.echo
+
+## Instructions for using GitHub tools
```
njhale, line 223: I think we use `START INSTRUCTIONS: <tool name(s)>` elsewhere

njhale (approved):

thedadams (approved):

g-linville (approved):

github/tool.gpt
drpebcak: looks like weve done that in just a couple places - jira, database, knowledge, and workspace files.

the format i used here is more common, but there is probably some value in standardizing on something.

PR #308, opened 2025-01-02
fix: echo the task's context

thedadams:
Since the context tool isn't using sys.echo, the prompt is fed to the LLM and the response of is being used as the context. This is not the intended behavior.

Using sys.echo here will mean that the text is used in the system prompt and the LLM is not used.

Issue: https://github.com/obot-platform/obot/issues/1066


njhale (approved):
Thanks for figuring this out!

StrongMonkey (approved):
💯

PR #314, opened 2025-01-06
chore: gitlab tools

drpebcak:
relates to https://github.com/obot-platform/obot/issues/866


gitlab/main.go
```diff
@@ -0,0 +1,254 @@
...
+	// Issues
+	case "queryIssues":
+		rawQuery := os.Getenv("ISSUE_QUERY")
+		var query map[string]any
+		if rawQuery != "" {
+			query, err = parseStringToMap(rawQuery)
+			if err != nil {
+				fmt.Printf("Failed to parse ISSUE_QUERY: %v\n", err)
+				os.Exit(1)
+			}
```
njhale, line 41: nit: if you factor this switch out into a separate function, you can return errors and have a single `fmt.Println(err); os.Exit(1);` in `main()`.

gitlab/tool.gpt
```diff
@@ -0,0 +1,227 @@
...
+Description: Approves a Merge Request.
+Share Context: GitLab Context
+Credential: ./credential
+Param: mr_id: The numeric internal ID of a Merge Request.
+Param: project_id: The numeric ID or url-escaped name of a Project.
+
+#!${GPTSCRIPT_TOOL_DIR}/bin/gptscript-go-tool approveMergeRequest
+
+---
+Name: Lookup User ID
```
njhale, line 196: nit: may also be nice to have a `Search Users` tool.

gitlab/tool.gpt
```diff
@@ -0,0 +1,227 @@
...
+Description: Takes a username and returns the ID for that user.
+Share Context: GitLab Context
+Credential: ./credential
+Param: gitlab_username: A GitLab username.
```
njhale, line 200: Do we expect this tool to get called a lot? If so, I'd consider optimizing it by making this tool a bulk operation that retrieves a list of user IDs given a set of usernames.

gitlab/tool.gpt
drpebcak: Perhaps, but it was not part of the scope in the original issue. `Lookup User ID` exists only to get the user ID for calls that require a numeric ID instead of accepting a username.

gitlab/tool.gpt
drpebcak: No, I don't think it would be called very often. Most operations accept a username, but there's one or two that require the numeric ID. Those sorts of calls wouldn't typically need to get a bunch of IDs at once.

gitlab/pkg/commands/issues.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return err
+	}
+	fmt.Printf("* Issue #%d: %s\n", issue.IID, issue.Title)
+	fmt.Printf("  Issue ID: %d\n", issue.ID)
+	fmt.Printf("  Project Name: %s\n", project.Name)
+	fmt.Printf("  Project ID: %d\n", issue.ProjectID)
+	fmt.Printf("  State: %s\n", issue.State)
+	fmt.Printf("  Author Username: %s\n", issue.Author.Username)
+	fmt.Printf("  Author ID: %d\n", issue.Author.ID)
```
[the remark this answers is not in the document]
drpebcak: Not sure we have an official stance, but this is how we have formatted the data in a few other places.

gitlab/.gitignore
thedadams, line 1: nit: these should be added to the top-level `.gitignore` file.

gitlab/pkg/commands/helpers.go
```diff
@@ -0,0 +1,11 @@
+package commands
+
+import "strings"
+
+func IndentString(s string) string {
+	lines := strings.Split(s, "\n")
+	for i, line := range lines {
+		lines[i] = "    " + line
+	}
+	return strings.Join(lines, "\n")
```
thedadams, line 10: Maybe this is slightly more efficient?
```suggestion
	return "    " + strings.ReplaceAll(s, "\n", "\n    ")
```

gitlab/pkg/commands/issues.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return err
+	}
+	if err := listIssueOutput(client, this); err != nil {
+		return err
+	}
+
+	notes, _, err := client.Notes.ListIssueNotes(projectID, issueIID, &gitlab.ListIssueNotesOptions{})
+	if err != nil {
+		log.Fatalf("Failed to fetch issue notes: %v", err)
```
thedadams, line 148: Can we just return this error and have the top-level error handling do it?

PR #315, opened 2025-01-06
feat: add DeepSeek model provider

sanjay920:
Add DeepSeek model provider to support DeepSeek models.
Ask me for API key details, i can give you one.

Depends on https://github.com/obot-platform/obot/pull/1128


deepseek-model-provider/server/server.go
```diff
@@ -0,0 +1,106 @@
...
+	req.URL.Host = "api.deepseek.com"
+	req.URL.Scheme = "https"
+	req.Host = req.URL.Host
+
+	if !strings.HasPrefix(req.URL.Path, "/v1") {
+		req.URL.Path = "/v1" + req.URL.Path
+	}
+
+	req.Header.Set("Authorization", "Bearer "+s.apiKey)
+}
```
njhale, line 106: Just making a note here that it seems like a lot of model providers have nearly identical logic. I think it may be worthwhile to, at some point, factor out most of this logic into an independent go module/package that we can inject with provider-specific logic.

For example, consider a `tools/model-provider/proxy` package with a `Run` function that accepts logic for tweaking request paths, bodies, and headers; transforming responses; etc.

deepseek-model-provider/tool.gpt
```diff
@@ -0,0 +1,11 @@
...
+Description: Model Provider for DeepSeek
+Metadata: envVars: OBOT_DEEPSEEK_MODEL_PROVIDER_API_KEY
+Model Provider: true
+Credential: ../model-provider-credential as deepseek-model-provider
+
+#!sys.daemon ${GPTSCRIPT_TOOL_DIR}/bin/gptscript-go-tool
+
+---
+!metadata:*:icon
+/admin/assets/deepseek-logo.svg
```
njhale, line 11: nit: add a space to the bottom of this file.

njhale (approved):
lgtm

deepseek-model-provider/server/server.go
sanjay920: agreed. future work :)

PR #316, opened 2025-01-06
enhance: add a tool to download content from a given url

njhale:
Add a `Download` tool to the `Browser` bundle that adds support for downloading
content from a given HTTP/S URL to a workspace file.

Addresses https://github.com/obot-platform/obot/issues/1132

### DIFF
--- a/browser/src/server.ts
+++ b/browser/src/server.ts
@@ -8,6 +8,7 @@ import { scrollToBottom } from './scrollToBottom.ts'
 import { randomBytes } from 'node:crypto'
 import { getSessionId, SessionManager } from './session.ts'
 import { screenshot, ScreenshotInfo } from './screenshot.ts'
+import { download } from './download.ts'
 
 async function main (): Promise<void> {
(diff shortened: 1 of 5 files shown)


browser/src/server.ts
```diff
@@ -25,6 +26,28 @@ async function main (): Promise<void> {
     res.send('OK')
   })
 
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  app.post('/download', async (req: Request, res: Response) => {
```
njhale, line 30: No need to start a browser session for this since we're using axios to get the content.

StrongMonkey (approved):

browser/src/server.ts
StrongMonkey: I think the problem is that you won't be able to get content behind javascript

njhale: I think this will primarily be used for downloading files and content from non-web pages. Maybe that's not true.

drpebcak (approved):

PR #319, opened 2025-01-07
fix: display outlook event dates and times in the users preferred timezone

njhale:
Addresses https://github.com/obot-platform/obot/issues/508 and https://github.com/obot-platform/obot/issues/556

Depends on https://github.com/obot-platform/tools/pull/321


StrongMonkey (approved):

drpebcak (approved):

outlook/calendar/tool.gpt
```diff
@@ -126,36 +126,32 @@ Param: response: The response to the invitation. Possible values are "accept", "
...
 Name: Outlook Calendar Context
 Type: context
-Share Context: Current Date and Time from ../../time
-Share Tools: Get Default Timezone
+Share Context: ../../time
 
 #!sys.echo
 
-## Instructions for using the Microsoft Outlook Calendar tools
+# START INSTRUCTIONS: Microsoft Outlook Calendar tools
```
drpebcak, line 136: I raised this elsewhere, but why change this? Most other places we are using the format this was in before.

njhale: I was under the impression that this was the more reliable way to wrap context text so the LLM will understand it. There's precedent elsewhere in the codebase. I'll undo this change since it seems to be contentious.

njhale: I dropped this change

drpebcak: looks like its still in here

njhale: hmm, I think I messed up a rebase here and lost the change. Will fix.

njhale: @drpebcak dropped for real this time.

PR #320, opened 2025-01-07
feat: add auth providers

g-linville:
This PR adds auth provider tools for Google and GitHub. This is to replace the built-in auth providers that are currently in the Obot gateway code, but will soon be refactored to use these instead.

Once refactored, Obot will reverse proxy traffic to these daemon tools, which will then either respond on their own (for the `/obot*` paths) or handle it with the oauth2proxy.

The `/obot-get-state` path is used to get information about the logged in user. Obot will send over a serialized request, which gets reconstructed into an `http.Request` here in the auth provider, so that we can extract the state from the cookie using the oauth2proxy library.


github-auth-provider/main.go
```diff
@@ -0,0 +1,207 @@
...
+	mux := http.NewServeMux()
+	mux.HandleFunc("/{$}", func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(fmt.Sprintf("http://127.0.0.1:%s", port)))
+	})
+	mux.HandleFunc("/obot-get-state", obotGetState(oauthProxy))
+	mux.HandleFunc("/obot-get-icon-url", obotGetIconURL)
+	mux.HandleFunc("/", oauthProxy.ServeHTTP)
+
+	fmt.Printf("listening on 127.0.0.1:%s\n", port)
+	if err := http.ListenAndServe("127.0.0.1:"+port, mux); err != nil {
```
thedadams, line 114: `ListenAndServe` always returns an error. I think we should catch the "server closed" error and not `exit 1` in that case.

github-auth-provider/main.go
```diff
@@ -0,0 +1,207 @@
...
+
+		reqObj, err := http.NewRequest(sr.Method, sr.URL, nil)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("failed to create request object: %v", err), http.StatusBadRequest)
+			return
+		}
+
+		for k, v := range sr.Header {
+			reqObj.Header[k] = v
+		}
```
thedadams, line 150: Does something like this work? Maybe a cast is needed?
```suggestion
		reqObj.Header = sr.Header
```

github-auth-provider/main.go
```diff
@@ -0,0 +1,207 @@
...
+	type response struct {
+		IconURL string `json:"iconURL"`
+	}
+
+	if err := json.NewEncoder(w).Encode(response{IconURL: iconURL}); err != nil {
+		http.Error(w, fmt.Sprintf("failed to encode response: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
thedadams, line 206: This is written the first time we write to `w`.

github-auth-provider/tool.gpt
```diff
@@ -0,0 +1,11 @@
+Name: GitHub
+Description: Auth provider for GitHub
+Metadata: envVars: OBOT_GITHUB_AUTH_PROVIDER_CLIENT_ID,OBOT_GITHUB_AUTH_PROVIDER_CLIENT_SECRET,OBOT_AUTH_PROVIDER_COOKIE_SECRET,OBOT_AUTH_PROVIDER_EMAIL_DOMAINS
+Metadata: optionalEnvVars: OBOT_GITHUB_AUTH_PROVIDER_TEAMS,OBOT_GITHUB_AUTH_PROVIDER_ORG,OBOT_GITHUB_AUTH_PROVIDER_REPO,OBOT_GITHUB_AUTH_PROVIDER_TOKEN,OBOT_GITHUB_AUTH_PROVIDER_ALLOW_USERS
+Credential: ../model-provider-credential as github-auth-provider
```
thedadams, line 5: Should we consider renaming this credential to something more descriptive?

google-auth-provider/pkg/env/env.go
```diff
@@ -0,0 +1,57 @@
+package env
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+)
+
+func LoadEnvForStruct[T any](s *T) error {
```
thedadams, line 10: Can we have only one of these?

google-auth-provider/main.go
```diff
@@ -0,0 +1,185 @@
...
+	type response struct {
+		IconURL string `json:"iconURL"`
+	}
+
+	if err := json.NewEncoder(w).Encode(response{IconURL: iconURL}); err != nil {
+		http.Error(w, fmt.Sprintf("failed to encode response: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
thedadams, line 184: Same comment about this already being written when we write `w`.

google-auth-provider/main.go
```diff
@@ -0,0 +1,185 @@
...
+	mux := http.NewServeMux()
+	mux.HandleFunc("/{$}", func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(fmt.Sprintf("http://127.0.0.1:%s", port)))
+	})
+	mux.HandleFunc("/obot-get-state", obotGetState(oauthProxy))
+	mux.HandleFunc("/obot-get-icon-url", obotGetIconURL)
+	mux.HandleFunc("/", oauthProxy.ServeHTTP)
+
+	fmt.Printf("listening on 127.0.0.1:%s\n", port)
+	if err := http.ListenAndServe("127.0.0.1:"+port, mux); err != nil {
```
thedadams, line 92: Catch "server closed" error.

thedadams (changes requested):
In addition to these comments, can we do a little more consolidation? Now that we are adding more Go-based model providers, there is an effort to combine as much of the common logic as is reasonable. Can we do the same thing here? Perhaps a common package at the top level?

github-auth-provider/main.go
g-linville: done

github-auth-provider/main.go
g-linville: done. it works

github-auth-provider/tool.gpt
g-linville: I renamed this to `placeholder-credential`

google-auth-provider/pkg/env/env.go
g-linville: I moved this into a common package

google-auth-provider/main.go
g-linville: done

g-linville:
@thedadams I created a common package to hold code that would be used by all auth providers. For now, it only includes the env for struct stuff.

g-linville:
Sorry for the force push. Needed to rebase to fix a conflict.

github-auth-provider/tool.gpt
```diff
@@ -0,0 +1,11 @@
+Name: GitHub
+Description: Auth provider for GitHub
+Metadata: envVars: OBOT_GITHUB_AUTH_PROVIDER_CLIENT_ID,OBOT_GITHUB_AUTH_PROVIDER_CLIENT_SECRET,OBOT_AUTH_PROVIDER_COOKIE_SECRET,OBOT_AUTH_PROVIDER_EMAIL_DOMAINS
+Metadata: optionalEnvVars: OBOT_GITHUB_AUTH_PROVIDER_TEAMS,OBOT_GITHUB_AUTH_PROVIDER_ORG,OBOT_GITHUB_AUTH_PROVIDER_REPO,OBOT_GITHUB_AUTH_PROVIDER_TOKEN,OBOT_GITHUB_AUTH_PROVIDER_ALLOW_USERS
+Credential: ../placeholder-credential as github-auth-provider
+
```
thedadams, line 6:
```suggestion
Metadata: noUserAuth: google-auth-provider

```

google-auth-provider/tool.gpt
```diff
@@ -0,0 +1,10 @@
+Name: Google
+Description: Auth provider for Google
+Metadata: envVars: OBOT_GOOGLE_AUTH_PROVIDER_CLIENT_ID,OBOT_GOOGLE_AUTH_PROVIDER_CLIENT_SECRET,OBOT_AUTH_PROVIDER_COOKIE_SECRET,OBOT_AUTH_PROVIDER_EMAIL_DOMAINS
+Credential: ../placeholder-credential as google-auth-provider
+
```
thedadams, line 5:
```suggestion
Metadata: noUserAuth: google-auth-provider

```

thedadams (changes requested):
All the model providers need this metadata field.

g-linville:
Force pushed to rebase on main and fix conflicts.

thedadams:
@g-linville Did you mean to close this?

g-linville:
Nope, lol. Not sure how I managed to close it.

google-auth-provider/main.go
```diff
@@ -0,0 +1,189 @@
...
+
+type SerializableState struct {
+	ExpiresOn         *time.Time `json:"expiresOn"`
+	AccessToken       string     `json:"accessToken"`
+	PreferredUsername string     `json:"preferredUsername"`
+	User              string     `json:"user"`
+	Email             string     `json:"email"`
+}
+
+func obotGetState(p *oauth2proxy.OAuthProxy) http.HandlerFunc {
```
thedadams, line 120: It would be great if this function could be imported from a shared package for all auth providers. This is very likely to be exactly the same for all auth providers.

I would be OK with this change being made when we add the next auth provider.

google-auth-provider/main.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+
+		if err := json.NewEncoder(w).Encode(ss); err != nil {
+			http.Error(w, fmt.Sprintf("failed to encode state: %v", err), http.StatusInternalServerError)
+			return
+		}
+	}
+}
+
+func obotGetIconURL(w http.ResponseWriter, r *http.Request) {
```
thedadams, line 162: Same comment here about importing this from a shared package.

thedadams (changes requested):
The newly added auth provider tools need the `Metadata: noUserAuth` directive added to their tools.

google-auth-provider/main.go
g-linville: Done

google-auth-provider/main.go
g-linville: Done

thedadams: Doesn't look like this was done.

g-linville: Oh I didn't pay close enough attention to this one lol. Sorry. I would like to keep this out of the common package, since the function is a bit different for our current two providers.

thedadams (approved):

PR #326, opened 2025-01-08
add validation for deepseek

sanjay920:


deepseek-model-provider/validate/main.go
```diff
@@ -0,0 +1,88 @@
...
+		var errResp ErrorResponse
+		if err := json.Unmarshal(body, &errResp); err == nil && errResp.Error.Message != "" {
+			return fmt.Errorf("%s", errResp.Error.Message)
+		}
+		return fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
+
+	var modelsResp ModelsResponse
+	if err := json.Unmarshal(body, &modelsResp); err != nil {
+		return fmt.Errorf("failed to parse response: %w", err)
```
iwilltry42, line 74: I wonder if we should log those (so admins can see them) but return a more generic error as that will be bubbled up all the way to the UI.

iwilltry42 (approved):

deepseek-model-provider/validate/main.go
sanjay920: good point. added this

deepseek-model-provider/tool.gpt
```diff
@@ -9,3 +9,10 @@ Credential: ../model-provider-credential as deepseek-model-provider
 ---
 !metadata:*:icon
 /admin/assets/deepseek-logo.svg
+
+---
+Name: validate
+Description: Validate the DeepSeek API key
+Metadata: envVars: OBOT_DEEPSEEK_MODEL_PROVIDER_API_KEY
+
+#!${GPTSCRIPT_TOOL_DIR}/validate/bin/gptscript-go-tool
```
thedadams, line 18: I don't think this works in general. For example, if you run Obot outside of the Docker container, then I believe this will fail. Specifically, it won't be able to find the `gptscript-go-tool` binary. The reason: GPTScript will only build the `main.go` file at the root of the repo; it won't look for nested `main.go` files. It work in Obot when you set the `OBOT_REGISTRY_URL` to a local directory and you have already run `make build` in the tools repo, then this will work.

I think the approach would be to have arguments: `server` and `validate` to the `main.go` file and run the respective code from there.

thedadams (changes requested):

deepseek-model-provider/tool.gpt
sanjay920: makes sense

deepseek-model-provider/main.go
```diff
@@ -1,19 +1,27 @@
...
+		validate.PrintError("OBOT_DEEPSEEK_MODEL_PROVIDER_API_KEY environment variable not set")
 		os.Exit(1)
 	}
 
+	args := os.Args[1:]
+	if len(args) == 1 && args[0] == "validate" {
+		if err := validate.Run(apiKey); err != nil {
+			validate.PrintError(err.Error())
+		}
+		os.Exit(0)
```
thedadams, line 22: Does `os.Exit(1)` make sense if there is an error?

deepseek-model-provider/validate/validate.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		logError("Failed to create request", fmt.Sprintf("error=%q", err))
+		return fmt.Errorf("failed to initialize validation")
+	}
+
+	req.Header.Set("Authorization", "Bearer "+apiKey)
+	req.Header.Set("Accept", "application/json")
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
```
thedadams, line 62: nit:
```suggestion
	resp, err := http.DefaultClient.Do(req)
```

thedadams (approved):
Looks good! Just a couple of small things

deepseek-model-provider/main.go
sanjay920: makes sense. added

PR #332, opened 2025-01-10
Refactor Model Providers to Use Shared Proxy Package + validation for OpenAI like providers

sanjay920:
This PR consolidates and standardizes the model provider implementations by introducing a shared proxy package. Key changes:

- Created a shared proxy package from OpenAI provider's server code
- Migrated **Groq, DeepSeek, xAI, Ollama, and vLLM** providers from their individual implementations to use the shared proxy
- Added consistent validation support across all relevant providers. (6 birds, 1 stone WRT https://github.com/obot-platform/obot/issues/1158)
- Removed duplicate server implementations
- Converted Python-based providers (Groq, Ollama) to Go
- Added model filtering support for provider-specific model types

The shared proxy package reduces code duplication, improves maintainability, and ensures consistent behavior across all model providers.


openai-model-provider/proxy/validate.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func logError(loggerPath string, msg string) {
+	log.Printf("time=%q level=error msg=%q logger=%s", time.Now().Format(time.RFC3339), msg, loggerPath)
+}
+
+func handleValidationError(loggerPath, msg string) error {
+	logError(loggerPath, msg)
+	fmt.Printf("{\"error\": \"%s\"}\n", msg)
+	return nil
```
iwilltry42, line 27: Should this always return `nil`? Looks like in the model providers you're checking if an error is being returned to do `os.Exit(1)` .

openai-model-provider/proxy/validate.go
```diff
@@ -0,0 +1,74 @@
...
+	"net/http"
+	"time"
+)
+
+func init() {
+	log.SetFlags(0)
+}
+
+func logInfo(loggerPath string, msg string) {
+	log.Printf("time=%q level=info msg=%q logger=%s", time.Now().Format(time.RFC3339), msg, loggerPath)
```
iwilltry42, line 17: Apart from the loggerPath this is not much different from the default log layout <time> <level> <msg> - why use a custom function here?
FWIW, `slog` is in the stdlib if you need structured logging (can e.g. enable json mode and use extra fields)

openai-model-provider/proxy/validate.go
iwilltry42, line 27: Should this always return `nil`? Looks like in the model providers you're checking if an error is being returned to do `os.Exit(1)` .

openai-model-provider/proxy/validate.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if resp.StatusCode != 200 {
+		return handleValidationError(loggerPath, invalidCredsMsg)
+	}
+
+	var modelsResp struct {
+		Object string `json:"object"`
+		Data   []any  `json:"data"`
+	}
+	if err := json.Unmarshal(body, &modelsResp); err != nil {
```
iwilltry42, line 64: You can use `json.NewDecoder` to decode right from the body (reader) into the struct without reading it first.

xai-model-provider/main.go
```diff
@@ -2,24 +2,65 @@ package main
...
-	"github.com/obot-platform/tools/xai-model-provider/server"
+	"github.com/obot-platform/tools/openai-model-provider/proxy"
 )
 
+const loggerPath = "/tools/xai-model-provider/validate"
+
+func ValidateXAIAPIKey(cfg *proxy.Config) error {
+	if cfg.APIKey == "" {
+		const msg = "Invalid xAI Credentials"
+		log.Printf("time=%q level=error msg=%q logger=%s", time.Now().Format(time.RFC3339), msg, loggerPath)
```
iwilltry42, line 19: Could as well export the log functions in the proxy package and re-use them here (if you don't want to switch to e.g. slog with a less custom format)

iwilltry42 (changes requested):
I like the logic consolidation and the refactoring - just have some confusion around the validation - it seems like it can never return an error (but I may be wrong here).

openai-model-provider/proxy/validate.go
iwilltry42: Actually it doesn't matter for the functionality of calling the validate tool in Obot since we're capturing this JSON-formatted error for both success and failure cases.

openai-model-provider/proxy/validate.go
sanjay920: fair. wasnt familiar with slog and was trying to match the servers log output format

openai-model-provider/proxy/proxy.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if cfg.RewriteModelsFn == nil {
+		cfg.RewriteModelsFn = DefaultRewriteModelsResponse
+	}
+
+	if cfg.ValidateFn != nil {
+		if err := cfg.ValidateFn(cfg); err != nil {
+			return fmt.Errorf("validation failed: %w", err)
+		}
```
iwilltry42, line 41: If I saw that correctly, all validation functions log the error but return `nil`, so this will never trigger (but likely should at this point?)

I think you can just return the errors directly in the validation functions and only in your validation func handle the error:

```go
func Validate(cfg *Config) error {
	if cfg.ValidateFn == nil {
		return nil
	}
	return handleValidationError(cfg.ValidateFn(cfg)) // this will be the only point calling handleValidationError... also, you can still log the error here if you want to (just need to figure out a way to get the loggerpath here or just include it in the err)
}
```

e.g. in the vLLM validation func you would then do 

```go
if cfg.APIKey == "" {
		return fmt.Errorf("missing vLLM APIKey") // text only needed if you're going to bubble it up
	}
```

iwilltry42 (changes requested):
Some nits and a suggestion for validation handling

openai-model-provider/proxy/validate.go
```diff
@@ -0,0 +1,52 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&modelsResp); err != nil {
+		return handleValidationError(loggerPath, invalidCredsMsg)
+	}
+
+	if len(modelsResp.Data) == 0 {
+		return handleValidationError(loggerPath, invalidCredsMsg)
+	}
+
+	slog.Info("Credentials are valid", "logger", loggerPath)
+	fmt.Println("{\"message\": \"Credentials are valid\"}")
```
thedadams, line 50: Do we need both of these logs?

sanjay920: I thought `fmt.Println("{\"message\": \"Credentials are valid\"}")` was necessary for the UI to pick up a notification but looks like that's wrong. removing one

ollama-model-provider/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/obot-platform/tools/ollama-model-provider
+
+go 1.23.4
+
+require (
+	github.com/obot-platform/tools/openai-model-provider v0.0.0
+)
+
+replace github.com/obot-platform/tools/openai-model-provider => ../openai-model-provider
```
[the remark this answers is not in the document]
sanjay920: this is a mistake on my end. was using for local dev

openai-model-provider/proxy/validate.go
```diff
@@ -35,22 +34,14 @@ func DoValidate(apiKey, urlStr, loggerPath, invalidCredsMsg string) error {
...
 	defer resp.Body.Close()
 
-	body, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return handleValidationError(loggerPath, invalidCredsMsg)
-	}
-
-	if resp.StatusCode != 200 {
-		return handleValidationError(loggerPath, invalidCredsMsg)
-	}
```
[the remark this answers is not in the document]
sanjay920: it's still there under `if resp.StatusCode != http.StatusOK {`

sanjay920:
The flow is:
1. Tool is invoked with `validate` argument
2. Config is created with provider-specific settings (host, API key, etc.)
3. `Validate()` is called which:
   - Makes HTTP request to provider's models endpoint
   - If any error occurs (bad credentials, network error, etc.):
     - Logs the error (for debugging)
     - Prints JSON-formatted error (for tool consumption)
     - Returns error to trigger exit(1)
4. Main program exits with status 1 if validation failed

openai-model-provider/proxy/validate.go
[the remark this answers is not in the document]
iwilltry42: Seems like it was gone in that one commit I reviewed :grimacing:

iwilltry42 (approved):

openai-model-provider/proxy/validate.go
```diff
@@ -0,0 +1,61 @@
...
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return handleValidationError(toolPath, "Invalid Response")
+	}
+
+	var modelsResp api.ModelsResponse
+	if err := json.Unmarshal(body, &modelsResp); err != nil {
+		return handleValidationError(toolPath, "Invalid Response Format")
+	}
```
thedadams, line 54: nit: so we don't have to process the body twice.
```suggestion
	var modelsResp api.ModelsResponse
	if err := json.NewDecoder(resp.Body).Decode(&modelsResp); err != nil {
		return handleValidationError(toolPath, "Invalid Response Format")
	}
```

thedadams (approved):

PR #334, opened 2025-01-10
enhance: use more descriptive names for database tools

njhale:
To reduce confusion in both the LLM and tool maintainers, rename the following tools in the `Database` bundle:

1. `Tables` -> `List Database Tables`
2. `Exec` -> `Exec Database Statement`
3. `Query` -> `Run Database Query`

Also fix tool descriptions to reflect the change from raw markdown output to JSON.

Addresses https://github.com/obot-platform/obot/issues/1208

Since these tools are in the `Capability` category, the "blast radius" of this change includes [the `TableHandler` implementation in obot](https://github.com/obot-platform/obot/blob/31ceab62d455bd798a069be034ea8effb1b41ac4/ui/user/src/lib/components/navbar/Tables.svelte#L8) and all tools/agents currently using them. That being the case, we
should hold-off on merging this change until the `TableHandler` implementation has been updated to the new tool names, or I keep duplicate tools that have the old names around until the handler implementation is updated.

### DIFF
--- a/database/tool.gpt
+++ b/database/tool.gpt
@@ -3,32 +3,32 @@ Name: Database
 Description: Tools for interacting with a database
 Metadata: category: Capability
 Metadata: icon: https://cdn.jsdelivr.net/npm/@phosphor-icons/core@2/assets/duotone/database-duotone.svg
-Share Tools: Query, Exec
+Share Tools: Run Database Query, Exec Database Statement
 
 ---
-Name: Tables
-Description: List all tables in the SQLite database and return the results in markdown format
+Name: List Database Tables
+Description: List all tables in the SQLite database and return a JSON object containing the results
 
-#!${GPTSCRIPT_TOOL_DIR}/bin/gptscript-go-tool listTables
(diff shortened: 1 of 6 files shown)


StrongMonkey (approved):

database/tool.gpt
```diff
@@ -3,32 +3,32 @@ Name: Database
...
 Metadata: category: Capability
 Metadata: icon: https://cdn.jsdelivr.net/npm/@phosphor-icons/core@2/assets/duotone/database-duotone.svg
-Share Tools: Query, Exec
+Share Tools: Run Query, Exec Statement
 
 ---
-Name: Tables
-Description: List all tables in the SQLite database and return the results in markdown format
+Name: List Tables
+Description: List all tables in the SQLite database and return a JSON object containing the results
```
cjellick, line 10: > "the results in markdown format"

Just to double check - you didnt change the return format right? You are just updating the description to match reality?

njhale: @cjellick Yes, I'm just updating it to match a change Darren made a while back. We were originally returning markdown formatted data. It's been JSON for a while at this point.

PR #341, opened 2025-01-14
feat: (anthropic-bedrock-model-provider) actually list Bedrock available models

iwilltry42:


anthropic-bedrock-model-provider/main.py
```diff
@@ -47,10 +50,36 @@ async def log_body(request: Request, call_next):
...
+            log(f"Failed to list models: {response}")
+            return JSONResponse(content={"error": f"Failed to list models - unexpected status code: {response}"},
+                                status_code=response["ReponseMetadata"]["HTTPStatusCode"])
+    except KeyError as e:
+        return JSONResponse(content={"error": f"Failed to list models - bad response: {e}"}, status_code=500)
+
+    try:
+        models = []
+        for model in response["modelSummaries"]:
+            if model["modelLifecycle"]["status"] != "ACTIVE" or "PROVISIONED" not in model["inferenceTypesSupported"]:
```
thedadams, line 72: `"PROVISIONED" not in model["inferenceTypesSupported"]` seems suspicious to me. I would expect `PROVISIONED` to be an inference type. If this is correct, then please ignore. Just wanted to call it out.

thedadams (approved):

anthropic-bedrock-model-provider/main.py
iwilltry42: There's PROVISIONED and ONDEMAND, but only the former worked properly without extra AWS configuration when I tested it, so I just ignore all others here.

iwilltry42:
Merging this as I need it for another PR :)

PR #354, opened 2025-01-20
Fix: Add visited url properly to memory store

StrongMonkey:
The url was not properly added to the memory store to restore scraping process. Also, record pdf files that have been scraped too. 

https://github.com/obot-platform/obot/issues/536

### DIFF
--- a/knowledge/data-sources/website/colly.go
+++ b/knowledge/data-sources/website/colly.go
@@ -25,16 +25,11 @@ const maxFileSize = 1024 * 1024 * 100
 
 func crawlColly(ctx context.Context, input *MetadataInput, output *MetadataOutput, logOut *logrus.Logger, gptscript *gptscript.GPTScript) error {
 	visited := make(map[string]struct{})
-	for url := range output.State.WebsiteCrawlingState.VisitedURLs {
-		filePath, err := convertUrlToFilePath(url)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


g-linville (approved):

thedadams (approved):

knowledge/data-sources/website/colly.go
```diff
@@ -212,13 +204,25 @@ func scrape(ctx context.Context, logOut *logrus.Logger, output *MetadataOutput,
 				}
 				linkURL = parsedLink
 			}
-			e.Request.Visit(linkURL.String())
+
+			if err := e.Request.Visit(linkURL.String()); err == nil {
```
StrongMonkey, line 208: Only marked url as visited if it has visited all the link under the url. Otherwise we could be missing url on the next restart.

knowledge/data-sources/website/main.go
```diff
@@ -32,8 +32,7 @@ type State struct {
 }
 
 type WebsiteCrawlingState struct {
-	CurrentURL  string              `json:"currentURL"`
-	VisitedURLs map[string]struct{} `json:"visitedURLs"`
+	VisitedURLs map[string]string `json:"visitedURLs"`
```
StrongMonkey, line 35: Changed it to a map of string and string to account for pdf file path.

g-linville (approved):

thedadams (approved):

PR #360, opened 2025-01-21
enhance: add create and write google docs tools

njhale:
Addresses https://github.com/obot-platform/obot/issues/1406

### DIFF
--- a/google/credential/tool.gpt
+++ b/google/credential/tool.gpt
@@ -6,7 +6,7 @@ Share Credential: ../../oauth2 as google
         openid
         https://www.googleapis.com/auth/userinfo.profile
         https://www.googleapis.com/auth/spreadsheets
-        https://www.googleapis.com/auth/documents.readonly
+        https://www.googleapis.com/auth/documents
         https://mail.google.com/
         https://www.googleapis.com/auth/gmail.readonly
         https://www.googleapis.com/auth/gmail.compose" as scope
--- a/google/docs/tool.gpt
+++ b/google/docs/tool.gpt
@@ -1,17 +1,37 @@
 ---
-Name: Google Docs 
+Name: Google Docs
 Metadata: bundle: true
 Description: Tools for managing Google Docs
-Share Tools: Read Doc
+Share Tools: Read Google Doc, Create Google Doc, Update Google Doc
 
 ---
-Name: Read Doc
-Description: Returns the content of a Google Doc in markdown format
+Name: Read Google Doc
+Description: Returns the content of a Google Doc in markdown format.
 Share Context: Google Docs Context
 Credential: ../credential
 Param: doc_ref: Google Docs ID or share link of the document to read
 
 #!/usr/bin/env python3 ${GPTSCRIPT_TOOL_DIR}/read_doc.py
+
+---
+Name: Update Google Doc
+Description: Updates a Google Doc
+Share Tools: Create Google Doc
+Share Context: Google Docs Context, Update Google Doc Context
+Credential: ../credential
+Param: doc_ref: Google Docs ID or share link of the document to read
+Param: new_doc_content: Markdown formatted content to replace the existing content of the document with
+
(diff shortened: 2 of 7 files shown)


google/credential/tool.gpt
```diff
@@ -7,6 +7,7 @@ Share Credential: ../../oauth2 as google
         https://www.googleapis.com/auth/userinfo.profile
         https://www.googleapis.com/auth/spreadsheets
         https://www.googleapis.com/auth/documents.readonly
```
drpebcak, line 9: Can you remove this scope? We don't need write AND readonly

google/docs/update_doc.py
```diff
@@ -0,0 +1,145 @@
...
+        # Update the current index to account for the added text
+        current_index += text_length
+
+    # Process elements in the markdown
+    for element in soup.contents:
+        if element.name in ['p']:
+            add_text_request(element.get_text() + "\n")
+        elif element.name in ['h1', 'h2', 'h3']:
+            # Apply bold style for headers
+            add_text_request(element.get_text() + "\n", bold=True)
```
njhale, line 68: This needs to change a bit. It's converting headers to bold instead of the respective google docs `hX` size.

tybalex:
Correct me if I'm wrong, but it looks like once the doc title is created, it can't get updated anymore by the `UpdateDoc` tool. Is this expected?

google/docs/tool.gpt
```diff
@@ -1,18 +1,38 @@
 ---
-Name: Google Docs 
+Name: Google Docs
 Metadata: bundle: true
 Description: Tools for managing Google Docs
-Share Tools: Read Doc
+Share Tools: Read Google Doc, Update Google Doc
```
thedadams, line 5:
```suggestion
Share Tools: Read Google Doc, Update Google Doc, Create Google Doc
```

google/docs/tool.gpt
```diff
@@ -1,18 +1,38 @@
...
+Param: new_doc_content: Markdown formatted content to replace the existing content of the document with
+
+#!/usr/bin/env python3 ${GPTSCRIPT_TOOL_DIR}/update_doc.py
+
+---
+Name: Create Google Doc
+Description: Creates a blank Google Doc
+Share Context: Google Docs Context
+Credential: ../credential
+Param: doc_title: The title of the document to create
```
thedadams, line 32: Any reason to not allow creating a document with content?

google/docs/update_doc.py
```diff
@@ -0,0 +1,145 @@
...
+                        }
+                    }
+                }
+            ]
+
+            # Issue a batch update request to clear the document
+            service.documents().batchUpdate(
+                documentId=file_id,
+                body={"requests": clear_requests}
+            ).execute()
```
thedadams, line 130: Can we make both of these batch requests at once?

google/docs/tool.gpt
njhale: I left this out on purpose because it's not really useful on its own.

google/docs/tool.gpt
njhale: They're two separate operations in the API. To me it seemed like dividing the responsibilities the same way results in simpler tools.

google/docs/update_doc.py
njhale: yeah, that's probably better. need to test if that causes issues with the doc indexing first.

google/docs/tool.gpt
thedadams: Interesting. Thanks for the explanation!

google/docs/tool.gpt
njhale: I added it to the bundle

njhale:
> Correct me if I'm wrong, but it looks like once the doc title is created, it can't get updated anymore by the UpdateDoc tool. Is this expected?

yeah, I decided to keep it simple. I can add the feature if we really need it.

StrongMonkey (approved):

thedadams (approved):

PR #361, opened 2025-01-22
feat: add support to use PATs instead of OAuth

thedadams:
Tools will now be able to specify, in their metadata, that their OAuth credential supports personal access tokens with the oauthPATSupported field being "true". If a tool does this, then the OAuth credential will prompt the user for a PAT instead of going through the OAuth flow.

This change includes this field on the GitHub and Atlassian tools.

Lastly, this change removes support for OAuth outside Obot. The OAuth credential tool will no longer look at the CLI config nor default to use the Gateway.

Issue: https://github.com/obot-platform/obot/issues/1372

### DIFF
--- a/atlassian/jira/tool.gpt
+++ b/atlassian/jira/tool.gpt
@@ -192,3 +192,7 @@ Jira
 ---
 !metadata:*:oauth
 atlassian
+
+---
+!metadata:*:supportsOAuthTokenPrompt
+true
--- a/atlassian/jira/tool.gpt
+++ b/atlassian/jira/tool.gpt
@@ -192,7 +192,3 @@ Jira
 ---
 !metadata:*:oauth
 atlassian
-
----
-!metadata:*:oauthPATSupported
-true
--- a/atlassian/jira/tool.gpt
+++ b/atlassian/jira/tool.gpt
@@ -192,3 +192,7 @@ Jira
 ---
 !metadata:*:oauth
 atlassian
+
+---
+!metadata:*:oauthPATSupported
+true
--- a/oauth2/main.go
+++ b/oauth2/main.go
@@ -11,7 +11,6 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"slices"
 	"strings"
 	"time"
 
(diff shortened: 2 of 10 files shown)


oauth2/main.go
```diff
@@ -347,3 +302,42 @@ func generateString() (string, error) {
...
+	}
+
+	g, err := gptscript.NewGPTScript(gptscript.GlobalOptions{})
+	if err != nil {
+		return fmt.Errorf("promptForPAT: failed to create GPTScript client: %w", err)
+	}
+	defer g.Close()
+
+	run, err := g.Run(context.Background(), "sys.prompt", gptscript.Options{
+		Input: fmt.Sprintf(`{"metadata": %s,"message":"Please enter your personal access token for %s.","fields":"Token","sensative": true}`, b, integration),
```
g-linville, line 325:
```suggestion
		Input: fmt.Sprintf(`{"metadata": %s,"message":"Please enter your personal access token for %s.","fields":"Token","sensitive": true}`, b, integration),
```

If this typo exists on the gptscript end, then it should be fixed there too.

g-linville (changes requested):

StrongMonkey (approved):

g-linville (approved):

thedadams:
Pushed a change in design (in a new commit), so requesting reviews.

atlassian/jira/tool.gpt
```diff
@@ -191,4 +191,8 @@ Jira
 
 ---
 !metadata:*:oauth
-atlassian
\ No newline at end of file
+atlassian
+
+---
+!metadata:*:supportsOAuthTokenPrompt
```
njhale, line 197: The description of this PR references an `oauthPATSupported` field. The field here is named differently. Is this intentional?

StrongMonkey: I think that's changed in this PR https://github.com/obot-platform/obot/pull/1387

StrongMonkey (approved):

njhale (approved):

PR #363, opened 2025-01-23
fix: change summarizer tool category to AI Summarization

tybalex:
Screenshot 2025-01-23 at 2 49 15 PM

one line fix to address https://github.com/obot-platform/obot/issues/1350#issuecomment-2608309095


sanjay920:
why are we making this change?

tybalex:
> why are we making this change?

Previously it was under the wrong category.

thedadams:
I would like to check to see if this is working as desired right now.

https://acorn-io.slack.com/archives/C03FU91U4SX/p1737634660199359

tybalex:
> I would like to check to see if this is working as desired right now.
> 
> https://acorn-io.slack.com/archives/C03FU91U4SX/p1737634660199359

Yes I agree with you. To me it probably makes more sense to keep this under the `Utility` category. 
Right now the issue is, the UI would display the summarizer tool as a sub-tool of time.

thedadams:
For now, we are going to fix this a different way. See https://github.com/obot-platform/tools/pull/368

PR #365, opened 2025-01-23
fix: auth providers: ensure cookie secrets are len 32

g-linville:
I'm still not sure exactly how it happened, but even though we generate 32 bytes for the cookie secret, I one time saw an auth provider crash on launch, saying it was only 31 bytes that it had. This is a little bandaid to prevent that from happening again. I saved the value that caused the problem so that I can investigate it more thoroughly in the future. For now, I just want to prevent it from happening again.

### DIFF
--- a/github-auth-provider/main.go
+++ b/github-auth-provider/main.go
@@ -44,6 +44,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	if len(cookieSecret) == 31 {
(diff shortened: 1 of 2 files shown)


cjellick (approved):

g-linville:
The real fix is here: https://github.com/obot-platform/obot/pull/1427

PR #366, opened 2025-01-23
Chore: Add openai o1 support

StrongMonkey:
Add openai o1 support in proxy to modify request body.


openai-model-provider/proxy/proxy.go
```diff
@@ -96,6 +101,41 @@ func (s *server) proxyDirector(req *http.Request) {
 	if s.cfg.PathPrefix != "" && !strings.HasPrefix(req.URL.Path, s.cfg.PathPrefix) {
 		req.URL.Path = s.cfg.PathPrefix + req.URL.Path
 	}
+
+	if req.Body == nil || req.Method != http.MethodPost {
```
thedadams, line 105: Does it make sense to do this?
```suggestion
	if req.Body == nil || s.cfg.UpstreamHost != "api.openai.com" || req.URL.Path != "/v1/chat/completion" {
```

g-linville: I like this, because it means that we can look for errors when we try to parse the body later.

StrongMonkey: yep, I will change this.

StrongMonkey:
@thedadams Pushed a change that make the response compatiable with stream client in gptscript.

openai-model-provider/proxy/proxy.go
```diff
@@ -96,6 +108,106 @@ func (s *server) proxyDirector(req *http.Request) {
...
+			resp.Body = io.NopCloser(bytes.NewBufferString(sseFormattedData))
+		} else {
+			resp.Body = io.NopCloser(bytes.NewBuffer(rawBody))
+		}
+	}
+
+	return nil
+}
+
+func isModelO1(model string) bool {
```
StrongMonkey, line 206: The current PR only supports o1, as o1-mini doesn't seem to support system message which could screw a lot of things. We could investigate in a different issue.

g-linville (approved):

openai-model-provider/proxy/proxy.go
```diff
@@ -96,6 +108,109 @@ func (s *server) proxyDirector(req *http.Request) {
...
+	req.Body = io.NopCloser(bytes.NewBuffer(modifiedBodyBytes))
+	req.ContentLength = int64(len(modifiedBodyBytes))
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Accept-Encoding", "")
+	req.Header.Set("Content-Type", "application/json")
+	return nil
+}
+
+func (s *server) modifyResponse(resp *http.Response) error {
+	if resp.StatusCode != http.StatusOK || resp.Body == nil || resp.Request.URL.Path != chatCompletionsPath || resp.Request.URL.Host != openaiBaseHostName {
```
thedadams, line 154: nit: `resp.Body` is always non-nil.
```suggestion
	if resp.StatusCode != http.StatusOK || resp.Request.URL.Path != chatCompletionsPath || resp.Request.URL.Host != openaiBaseHostName {
```

openai-model-provider/proxy/proxy.go
```diff
@@ -96,6 +108,109 @@ func (s *server) proxyDirector(req *http.Request) {
...
+		return nil
+	}
+
+	if resp.Header.Get("Content-Type") == "application/json" {
+		var respBody openai.ChatCompletionResponse
+		rawBody, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("failed to read response body: %w", err)
+		}
+		if err := json.Unmarshal(rawBody, &respBody); err == nil && isModelO1(respBody.Model) {
```
thedadams, line 164: So we only have to process the body once.
```suggestion
		if err := json.Decode(resp.Body).Decode(&respBody); err == nil && isModelO1(respBody.Model) {
```

openai-model-provider/proxy/proxy.go
```diff
@@ -96,6 +108,109 @@ func (s *server) proxyDirector(req *http.Request) {
...
+	return nil
+}
+
+func (s *server) modifyResponse(resp *http.Response) error {
+	if resp.StatusCode != http.StatusOK || resp.Body == nil || resp.Request.URL.Path != chatCompletionsPath || resp.Request.URL.Host != openaiBaseHostName {
+		return nil
+	}
+
+	if resp.Header.Get("Content-Type") == "application/json" {
+		var respBody openai.ChatCompletionResponse
```
thedadams, line 159: Have to close the body if we are reading and replacing it. I think you need to save the original body in a variable so you don't close the new body.
```suggestion
		originalBody := resp.Body
		defer originalBody.Close()
		var respBody openai.ChatCompletionResponse
```

openai-model-provider/proxy/proxy.go
```diff
@@ -96,6 +108,109 @@ func (s *server) proxyDirector(req *http.Request) {
...
+								Content:      choice.Message.Content,
+								Role:         choice.Message.Role,
+								FunctionCall: choice.Message.FunctionCall,
+								ToolCalls:    choice.Message.ToolCalls,
+							},
+							FinishReason: choice.FinishReason,
+						})
+					}
+					return choices
+				}(),
```
thedadams, line 187: Whoa! I've never seen this before. I think I like it.

openai-model-provider/proxy/proxy.go
```diff
@@ -63,7 +74,8 @@ func Run(cfg *Config) error {
 		ModifyResponse: cfg.RewriteModelsFn,
 	})
 	mux.Handle("/v1/", &httputil.ReverseProxy{
-		Director: s.proxyDirector,
+		Director:       s.proxyDirector,
+		ModifyResponse: s.modifyResponse,
```
thedadams, line 78: I think it would be good if we could find a way to not put this on every model provider that uses this proxy code, and only put it on the OpenAI model provider.

thedadams (changes requested):

openai-model-provider/proxy/proxy.go
StrongMonkey: yeah, probably better to just close the original body without using defer

openai-model-provider/proxy/proxy.go
StrongMonkey: yes, it would need us to put the common code into a seperate folder/repo with its own go modules. Right now if I can update dep for openai-model-provider it pulls other in.

StrongMonkey: wait, actually, yeah I didn't realize the proxy code is shared. Let me think..

StrongMonkey: ok, thorsten's custom path comes right on time. This switches to use that.

openai-model-provider/proxy/proxy.go
```diff
@@ -87,25 +87,31 @@ func Run(cfg *Config) error {
...
-	for path, handler := range cfg.CustomPathHandleFuncs {
-		mux.HandleFunc(path, handler)
-	}
-
 	// Register default handlers only if they are not already registered
 	if _, exists := cfg.CustomPathHandleFuncs["/{$}"]; !exists {
 		mux.HandleFunc("/{$}", s.healthz)
 	}
-	if _, exists := cfg.CustomPathHandleFuncs["/v1/models"]; !exists {
+	if handler, exists := cfg.CustomPathHandleFuncs["/v1/models"]; !exists {
```
StrongMonkey, line 94: @iwilltry42 can you review this?

iwilltry42: LGTM!
I don't like all those if-else statements, but have no cleaner solution right now (unfortunately, mux doesn't take care of the path sorting and neither allows overrides nor looking up what paths are already handled.

iwilltry42 (approved):

openai-model-provider/proxy/proxy.go
```diff
@@ -9,8 +9,14 @@ import (
...
 
+var (
+	OpenaiBaseHostName = "api.openai.com"
+
+	ChatCompletionsPath = "/v1/chat/completions"
+)
+
 type Config struct {
-	url *url.URL
+	Url *url.URL
```
thedadams, line 19: nit:
```suggestion
	URL *url.URL
```

thedadams (approved):

iwilltry42:
@pixeye33 that would only make sense, if one were to use the generic provider to connect to OpenAI directly instead of using the OpenAI provider, right?
Or is there an o1 variant with the same limitations that you can run in e.g. Ollama?

Now that I think about it - it may be needed for Azure OpenAI which has access to o1 as well.

PR #367, opened 2025-01-24
enhance: enable js for google search tool

njhale:
Enable Javascript when gathering page content in the `Google Search` tool. This allows more content to be extracted from web pages that require Javascript to load properly.

Addresses part of https://github.com/obot-platform/obot/issues/1423

### DIFF
--- a/google/search/src/search.ts
+++ b/google/search/src/search.ts
@@ -24,23 +24,13 @@ export async function search (
 
   const encodedQuery = encodeURIComponent(query)
   const searchUrl = `https://www.google.com/search?q=${encodedQuery}&udm=14`
-  
+
   const foundURLs = new Set<string>()
   const results: Array<Promise<SearchResult | null>> = []
 
   const page = await context.newPage()
-  const noJSPages = await Promise.all(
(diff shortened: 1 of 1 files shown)


google/search/src/search.ts
```diff
@@ -24,21 +24,14 @@ export async function search (
...
   const searchUrl = `https://www.google.com/search?q=${encodedQuery}&udm=14`
-  
+
   const foundURLs = new Set<string>()
   const results: Array<Promise<SearchResult | null>> = []
 
   const page = await context.newPage()
-  const noJSPages = await Promise.all(
+  const pages = await Promise.all(
     Array.from({ length: maxResults }, async () => {
```
ryanhopperlowe, line 33: nit: you can remove async/await from this since you're returning a promise

```ts
const pages = await Promise.all(
    Array.from({ length: maxResults }, () => context.newPage())
)
```

ryanhopperlowe (approved):
1 nit, but LGTM

google/search/src/search.ts
njhale: Thanks for the callout here. Fixed!

PR #368, opened 2025-01-24
chore: unbundle the time tool

thedadams:
Having the time tool bundled causes issues with other "utilities." Also, the time tool is just one tool, so we can unbudle it for now.


thedadams:
Issue: https://github.com/obot-platform/obot/issues/1350

StrongMonkey (approved):

PR #375, opened 2025-01-28
enhance: openai-model-provider proxy - combine options into single baseURL and add customPathHandleFuncs for customizability

iwilltry42:
E.g. Google Gemini Vertex AI has OpenAI API compatibility, but not for everything. /embeddings and /models are not supported, but still we gain the benefit of having /chat/completions and not having to translate the completion requests/responses.
Also, Gemini as the example at hand doesn't use the `/v1` API base.

With the `baseURL` setting we are closer to the settings we have/had in GPTScript and Knowledge

### DIFF
--- a/openai-model-provider/proxy/proxy.go
+++ b/openai-model-provider/proxy/proxy.go
@@ -5,44 +5,69 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httputil"
+	"net/url"
 	"strings"
 )
 
 type Config struct {
-	APIKey          string
-	Port            string
-	UpstreamHost    string
-	UseTLS          bool
-	ValidateFn      func(cfg *Config) error
+	url *url.URL
+
+	// ListenPort is the port the proxy server listens on
+	ListenPort string
+
+	// Name is the name of the provider, used for logging
+	Name string
+
+	// BaseURL is the upstream model API URL, e.g. "https://api.openai.com/v1" - MUST include the basePath, if any (e.g. /v1)
+	BaseURL string
+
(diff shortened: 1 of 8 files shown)


openai-model-provider/proxy/proxy.go
```diff
@@ -5,44 +5,69 @@ import (
...
 
 type server struct {
 	cfg *Config
 }
 
-func Run(cfg *Config) error {
-	if cfg.Port == "" {
-		cfg.Port = "8000"
+func (cfg *Config) ensureURL() error {
+
```
thedadams, line 42: nit:
```suggestion
```

openai-model-provider/proxy/proxy.go
```diff
@@ -5,44 +5,69 @@ import (
...
 	}
 
-	// Remove any trailing slashes from UpstreamHost
-	cfg.UpstreamHost = strings.TrimRight(cfg.UpstreamHost, "/")
+	cfg.url = u
+	return nil
+}
+
+func Run(cfg *Config) error {
+
```
thedadams, line 67: nit:
```suggestion
```

openai-model-provider/proxy/validate.go
```diff
@@ -16,14 +16,14 @@ func handleValidationError(loggerPath, msg string) error {
 }
 
 func (cfg *Config) Validate(toolPath string) error {
-	scheme := "https"
-	if !cfg.UseTLS {
-		scheme = "http"
+
```
thedadams, line 19: nit:
```suggestion
```

thedadams (approved):

g-linville (approved):

PR #379, opened 2025-01-28
enhance: add ability to create and move google docs to specific drive folders

njhale:
Addresses https://github.com/obot-platform/obot/issues/1491

### DIFF
--- a/google/credential/tool.gpt
+++ b/google/credential/tool.gpt
@@ -9,5 +9,6 @@ Share Credential: ../../oauth2 as google
         https://www.googleapis.com/auth/documents
         https://mail.google.com/
         https://www.googleapis.com/auth/gmail.readonly
-        https://www.googleapis.com/auth/gmail.compose" as scope
+        https://www.googleapis.com/auth/gmail.compose
+        https://www.googleapis.com/auth/drive" as scope
 Type: credential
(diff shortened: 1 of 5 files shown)


google/credential/tool.gpt
```diff
@@ -9,5 +9,7 @@ Share Credential: ../../oauth2 as google
         https://www.googleapis.com/auth/documents
         https://mail.google.com/
         https://www.googleapis.com/auth/gmail.readonly
-        https://www.googleapis.com/auth/gmail.compose" as scope
+        https://www.googleapis.com/auth/gmail.compose
+        https://www.googleapis.com/auth/drive
+        https://www.googleapis.com/auth/drive.file" as scope
```
drpebcak, line 14: Why both `drive` and `drive.file`? `drive` is much more permissive.. so you wouldn't need both.

`drive.file` currently would not work very well with obot, because we don't have a workflow for changing the files you have given access to.

drpebcak (changes requested):

google/docs/tool.gpt
```diff
@@ -21,6 +21,7 @@ Share Context: Google Docs Context, Update Google Doc Context
 Credential: ../credential
 Param: doc_ref: Google Docs ID or share link of the document to read
 Param: new_doc_content: Markdown formatted content to replace the existing content of the document with
+Param: new_drive_dir: Optional folder path in Google Drive to move the document to after updating it. Use "/" to move the document back to the root folder.
```
StrongMonkey, line 24: is there a reason why this is called `new_drive_dir` while the other is called `doc_drive_dir`?

njhale: it follows the convention set by the `new_doc_content` parameter. it's meant to give the LLM more of a hint that it's an update and not for locating the file.

g-linville (approved):

drpebcak (approved):

PR #383, opened 2025-01-28
chore: hubspot tools

drpebcak:

### DIFF
--- a/hubspot/crm/tool.gpt
+++ b/hubspot/crm/tool.gpt
@@ -1,0 +1,54 @@
+Name: HubSpot CRM
+Description: Tools for working with HubSpot CRM
+Metadata: bundle: true
+Share Context: HubSpot CRM Context
+Share Tools: ./associations.gpt, ./companies.gpt, ./contacts.gpt, ./deals.gpt, lists.gpt, owners.gpt, ./properties.gpt, ./tickets.gpt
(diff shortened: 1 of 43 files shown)


hubspot/crm/tool.gpt
```diff
@@ -0,0 +1,54 @@
...
+---
+name: Get Unix Timestamp
+description: Get the current Unix timestamp in milliseconds
+
+#!/usr/bin/env python3
+
+import time
+
+unix_timestamp_ms = round(time.time() * 1000)
+print(f"The current unix timestamp in milliseconds is {unix_timestamp_ms}")
```
thedadams, line 50: Not sure if this tool was ported over from somewhere else, but are we going to run into timezone issues with this? Should this be in the `Time` tool and use the timezone stuff added there?

thedadams (approved):
Just one question, everything else looks good.

hubspot/crm/tool.gpt
drpebcak: It was ported over from the original version of these tools, but a unix timestamp doesnt have a timezone anyway so it shouldnt cause any problems. arguably it could be part of the time tool, but its also somewhat specific to hubspot, because they want the unix timestamp in milliseconds, where normally an unix timestamp is in seconds

g-linville (approved):
Lgtm, all my comments are just little changes that don't really impact functionality, plus one question

PR #396, opened 2025-01-31
fix: modify Jira tool for API key support

thedadams:
The Jira API is different for OAuth versus API keys. This change modifies the Jira tool so that it works with API keys and OAuth.

Issue: https://github.com/obot-platform/obot/issues/1555

### DIFF
--- a/atlassian/jira/index.js
+++ b/atlassian/jira/index.js
@@ -16,21 +16,38 @@ const command = process.argv[2]
 async function main() {
 
     try {
-        if (!process.env.ATLASSIAN_OAUTH_TOKEN) {
-            throw new Error("ATLASSIAN_OAUTH_TOKEN is required")
+        if (!process.env.ATLASSIAN_OAUTH_TOKEN && (!process.env.ATLASSIAN_API_TOKEN || !process.env.ATLASSIAN_EMAIL || !process.env.ATLASSIAN_SITE_URL)) {
+            throw new Error("ATLASSIAN_OAUTH_TOKEN or all of ATLASSIAN_API_TOKEN, ATLASSIAN_EMAIL, and ATLASSIAN_SITE_URL are required")
+        }
+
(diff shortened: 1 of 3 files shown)


atlassian/jira/index.js
```diff
@@ -16,28 +16,49 @@ const command = process.argv[2]
...
+                'Authorization': `${authType} ${authToken}`,
                 'Accept': 'application/json',
             },
         })
 
         let result = null
         switch (command) {
             case "listJiraSites":
+                if (process.env.ATLASSIAN_SITE_URL) {
+                    console.log("listJiraSites is not needed when ATLASSIAN_SITE_URL is set, so you can continue without needing the site_id argument")
```
njhale, line 59: Could we just return the first component of the subdomain here instead? The side_id argument is ignored when a PAT credential is used, so it won't matter if it's not a real Jira ID, and we won't need to add the additional line context (and hope the LLM will respect it).

IMO, the only thing that could interfere with that is if the LLM has info about Jira site IDs in its training set and gets confused by the dummy ID. Although, I think it's pretty unlikely given how explicit our tool definitions and context is wrt site_id usage.

njhale (changes requested):

atlassian/jira/index.js
thedadams: Does that provide some value?

njhale (approved):
I tested this out and it works well for OAuth and the PAT, so I won't block this PR on my earlier suggestion, but I think it's worth considering as an alternative if folks share the same apprehension about the reliability of context-based solutions.

atlassian/jira/index.js
njhale: Just that it depends less on the LLM doing the right thing based on the context. I've been bitten by that before, especially when a bunch of tools are added to an agent and the context grows large.

PR #397, opened 2025-02-03
fix: Fix LinkedIn get_user() and create_post() method.

tybalex:
**What does this PR do:**
- address issue https://github.com/obot-platform/obot/issues/1567 by adding a retry mechanism to the `get_user()` function.
- address issue https://github.com/obot-platform/obot/issues/1576 by providing a manually crafted link to the post created.

### DIFF
--- a/linkedin/tools/posts.py
+++ b/linkedin/tools/posts.py
@@ -5,7 +5,7 @@ import os
 import requests
 
 
-async def create_post(client: RestliClient):
+async def create_post(client: RestliClient) -> dict:
     r = get_user(client)
     user_id = r["sub"]
     content = os.getenv("CONTENT")
@@ -96,10 +96,14 @@ async def create_post(client: RestliClient):
         raise ValueError(
             f"Error: failed to create post. Status code: {response.status_code}. Response: {response.entity}"
         )
+
+    response.entity["link"] = (
+        f"https://www.linkedin.com/feed/update/{response.entity['id']}"  # craft the link manually
+    )
     return response.entity
 
 
-def _register_upload(client: RestliClient, category: str):
+def _register_upload(client: RestliClient, category: str) -> tuple[str, dict]:
     url = "https://api.linkedin.com/v2/assets?action=registerUpload"
 
     headers = {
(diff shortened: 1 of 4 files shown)


linkedin/tools/posts.py
```diff
@@ -97,10 +96,14 @@ async def create_post(client: RestliClient):
         raise ValueError(
             f"Error: failed to create post. Status code: {response.status_code}. Response: {response.entity}"
         )
+
+    response.entity["link"] = (
+        f"https://www.linkedin.com/feed/update/{response.entity['id']}"  # craft the link manually
```
tybalex, line 101: the link to post created

linkedin/tools/users.py
```diff
@@ -1,8 +1,30 @@
...
+import time
+import sys
 
 
-@tool_registry.register_tool("GetCurrentUser")
-def get_user(client: RestliClient):
-    response = client.get(resource_path="/userinfo", access_token=ACCESS_TOKEN)
+def get_user(client: RestliClient) -> dict:
+
+    MAX_RETRIES = 3
```
tybalex, line 9: 3 retries to prevent a weird 401 error from the Linkedin API side.

linkedin/tools/users.py
```diff
@@ -1,8 +1,30 @@
...
+    DELAY = 1  # seconds between retries
+
+    for attempt in range(1, MAX_RETRIES + 1):
```
thedadams, line 12: It seems strange to me that we would need to retry this, Especially for a 401 error.

thedadams (approved):

linkedin/tools/users.py
tybalex: yeah, I am totally with you. Here is the detail of the issue:
Sometimes, the first time the tool makes a `get_user()` request , it gets such error response back:
```
\"status\": 401, \"serviceerrorcode\": 65601, \"code\": \"revoked_access_token\
```
and then the second time it just works without changing anything.
==================
My theory is: sometimes, the certain part of the Oauth process in the LinkedIn API server is so slow, such that after the access token is granted, our `get_user()` request hits the server earlier than the moment the access token is marked as a valid one in their system.

So I did find another way to fix this : add something like `time.sleep(2)` to the beginning of this function, and it works ..But  a `retry` probably makes more sense.

PR #398, opened 2025-02-03
Feat: General File Reader Tool

tybalex:
This PR adds a general `File-Reader Tool` by running knowledge load tool with python gptscript SDK.  
This tool can parse the content of input workspace file(word, pdfs, slides) and returns to the LLM. 
Maybe we can replace `workspace_read` with this in the future..?


iwilltry42 (approved):

tybalex:
> Hm. This isn't really a file reader tool. It's a "Convert to markdown" tool. I'm not sure that's what we want.

Fair, probably not markdown, parsing the content to plain text would be enough. I guess markdown format is more readable?

> I had envisioned a tool that reads a file and sends the contents directly to the LLM. Do you not think that approach will work well?

We can support both, and make it sends content back to llm by default. In the case of when the file has too much text, then it should probably write to a file and use a summarizer tool to handle the text.

> What happens if the user just says "Read the court filings file" or "What's the highest grossing office in my spreadsheet?" Will this tool be called?

For structured data like a spreadsheet, we will need to handle it separately.

tybalex:
I made an update, now this tool will send the content directly to LLM by default. If user specify an output file to save to, then it will do so.

knowledge/file-reader.gpt
```diff
@@ -0,0 +1,26 @@
+Name: File Reader
+Description: This tool reads any supported type of text-heavy documents(PDF, DOCX, PPTX, HTML) in the workspace, converts its content to markdown format. The parsed content will be send back to LLM or saved to a workspace file if output_file is provided.
```
cjellick, line 2:
```suggestion
Description: This tool reads any supported type of text-heavy documents(PDF, DOCX, PPTX, HTML) in the workspace, converts its content to markdown format. The parsed content will be sent back to LLM or saved to a workspace file if output_file is provided.
```

cloudnautique:
@tybalex what happens if the amount of text exceeds the context window? Is it something we handle gracefully?

tybalex:
Code updated:
- Use python GPTScript SDK to run knowledge load
- handle long text (I arbitrarily set it to 10k tokens) by integrating with the summarizer.

Maybe we can replace `workspace_read` with this in the future..?

tybalex:
Close this one and will open a new one. The file-reader-tool will be merged to workspace-read

PR #399, opened 2025-02-03
chore: rename linkedin bundle

njhale:
- Rename the `LinkedIn` bundle to `LinkedIn Publishing` to clarify its
  usage
- Make tool names, bundle description, and context more explicit

Addresses https://github.com/obot-platform/obot/issues/1583


iwilltry42 (approved):

index.yaml
```diff
@@ -25,8 +25,8 @@ tools:
   slack:
     reference: ./slack
     all: true
-  linkedin:
-    reference: ./linkedin
+  linkedin-publishing:
+    reference: ./linkedin-publishing
```
StrongMonkey, line 29: Nit:  it would be better organzied if we put folder like ./linkedin/publishing so it other linkedin tools can go to subfolder.

StrongMonkey (approved):

PR #402, opened 2025-02-04
fix: convert slack message datetimes to user tz in code

njhale:
- Convert slack message datetimes to user tz in code
- Improve slack search messages context

Addresses https://github.com/obot-platform/obot/issues/584 again.

### DIFF
--- a/slack/src/tools.js
+++ b/slack/src/tools.js
@@ -8,7 +8,6 @@ export async function userContext(webClient) {
     console.log(`Real Name: ${userResult.user.profile.real_name}`)
     console.log(`Display Name: ${userResult.user.profile.display_name}`)
     console.log(`User ID: ${result.user_id}`)
-    console.log(`Default time zone: ${userResult.user.tz_label} (UTC offset ${userResult.user.tz_offset / 3600} hours)`)
 }
 
 export async function listChannels(webClient) {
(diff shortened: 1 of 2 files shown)


slack/src/tools.js
```diff
@@ -403,6 +402,20 @@ function userToString(user) {
     return str
 }
 
+const userTimezone = (() => {
+    const envTz = (process.env.OBOT_USER_TIMEZONE ?? 'UTC').trim();
+
+    // Verify the time zone or default to UTC
+    let timeZone = envTz;
+    try {
+        new Intl.DateTimeFormat(undefined, { timeZone }).format();
```
g-linville, line 411: Does this need to be assigned back to something? Or does it modify the `timeZone` variable?

slack/tool.gpt
```diff
@@ -198,10 +198,28 @@ When mentioning a user in a message you create, use the format <@USERID>, includ
...
+- Boolean Operators: Use `AND`, `OR`, and `NOT` to refine searches.
+  Examples:
+  - `project AND update` finds messages containing both "project" and "update".
+  - `project OR update` finds messages containing either "project" or "update".
+  - `project NOT update` finds messages containing "project" but excluding "update".
+- Combining multiple modifiers: Separate each modifier and search term with spaces (order does not matter). Example: `update from:@alice in:#general after:2025-01-01 -in:#random`
+
+The following search modifiers can be used in search-messages `query` arguments:
+- `after:<date>`: Matches messages sent after the specified date (exclusive, so excludes messages from the given date). Must be in `YYYY-MM-DD` format (no time/timezone). Example: `after:2025-01-01`
+- `before:<date>`: Matches messages sent before the specified date (exclusive, so excludes messages from the given date). Example: `before:2025-01-01`
```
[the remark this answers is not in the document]
g-linville: Imo we should keep them. I don't think that that potential mismatch is a big deal.

slack/src/tools.js
njhale: It's just verifying the timezone string from the environment variable is valid. If it isn't, an exception is thrown and it defaults back to UTC.

g-linville: Oh, I see. Thanks.

g-linville (approved):

drpebcak (approved):

PR #403, opened 2025-02-04
Tool index refactoring

StrongMonkey:
This pr removes category and `all: true` per https://github.com/obot-platform/obot/issues/1026.

The import logic will be changed in obot to accommodate it.

It doesn't rely on bundle and category metadata anymore. Category can be used to defined bigger group other than bundle.

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -1,68 +1,49 @@
 tools:
   sendgrid:
     reference: ./sendgrid
-    all: true
   file-summarizer:
     reference: ./file-summarizer
-    all: true
   database:
     reference: ./database
   zoom:
     reference: ./zoom
-    all: true
   atlassian-jira:
     reference: ./atlassian/jira
-    all: true
   wordpress:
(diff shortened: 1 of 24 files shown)


index.yaml
```diff
@@ -1,68 +1,49 @@
...
-    all: true
   outlook-calender:
     reference: ./outlook/calendar
-    all: true
   github:
     reference: ./github
-    all: true
   time:
     reference: ./time
+    nameOverride: time
```
StrongMonkey, line 46: This is something I added to workaround a migration issue that has to deal with tool name change. Normally. under the current logic, if a tool is a bundle it has be to defined as the name `x-bundle`. This breaks the previous naming conversion it had with old agent/workflow. So this is just a way to override the name to not follow the same conversion.

tybalex (approved):

thedadams (approved):

njhale (approved):

PR #410, opened 2025-02-05
fix: google search workspace id extraction and session management

njhale:
Properly extract the thread workspace ID from the `X-GPTSCRIPT-ENV` header.

Before this change, the tool would fail to extract it and would fallback on the value set in its environment variables. Since the session ID -- used for managing sessions -- is derived from the extracted workspace ID and the value in the daemon tool's env is static, every thread ended up sharing the same session.

Additionally, redesign session management to:
- use a third-party TTL cache package to handle session expiration and
  cleanup
- use session scoped locks to reduce contention when provisioning new
  sessions
- remove potential race conditions

Addresses:
- https://github.com/obot-platform/obot/issues/1651
- https://github.com/obot-platform/obot/issues/1653

### DIFF
--- a/google/search/src/session.ts
+++ b/google/search/src/session.ts
@@ -5,7 +5,8 @@ import { type IncomingHttpHeaders } from 'node:http'
 import { createHash } from 'node:crypto'
 import { type BrowserContext } from 'playwright'
 import { newBrowserContext } from './context.ts'
-import { Mutex } from 'async-mutex'
+import TTLCache from '@isaacs/ttlcache'
+import AsyncLock from 'async-lock'
 
 const APP_CACHE_DIR = (() => {
   const homeDir = os.homedir()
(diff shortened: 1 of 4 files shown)


google/search/src/session.ts
```diff
@@ -34,117 +35,163 @@ async function clearAppCacheDir (): Promise<void> {
...
-    }
 
+  for (const env of envArray) {
+    if (env == null) continue
     for (const pair of env.split(',')) {
-      const [key, value] = pair.split('=')
-      if (key === 'GPTSCRIPT_WORKSPACE_ID') {
-        return value ?? process.env.GPTSCRIPT_WORKSPACE_ID
-      }
+      const [key, value] = pair.split('=').map((part) => part.trim())
```
StrongMonkey, line 192: oh, is the space that were causing the problem?

StrongMonkey (approved):

google/search/src/session.ts
njhale: heh, yup

thedadams (approved):

PR #413, opened 2025-02-05
feat: onedrive word write tool

iwilltry42:
Adds a new `Write Doc` tool that allows for creating new docx documents or updating existing ones in OneDrive.
It works with OneDrive item IDs or filepaths within OneDrive, creating directories as needed.
Conversion from Markdown to Docx is done either via Pandoc or LibreOffice, whichever is available.
In our docker images we definitely have LibreOffice (soffice binary) available.

Example of a Doc after:
- having the AI create a writing guide
- tool call to create a file writingguide.docx in root
- tool call to add some more content to it (third section)
- tool call to write that same content to a file in a nested folder

Screenshot 2025-02-05 at 17-06-49 mynewdoc docx

Issue https://github.com/obot-platform/obot/issues/1468

Note: It may make sense to also rely on pandoc/soffice instead of docconv to transform back from docx to markdown when reading a file.


word/pkg/graph/docs.go
```diff
@@ -17,13 +24,220 @@ func (d DocInfo) String() string {
...
+	}
+
+	var doc graphmodels.DriveItemable
+	var err error
+
+	// Check if the file exists by path or ID
+	if strings.ContainsAny(filename, "/.:_-") {
+		doc, err = getItemByPath(ctx, client, driveID, filename)
+		if err != nil {
+			slog.Info("Failed to get item by path. It may not exist so we create it.", "path", filename, "error", err)
```
g-linville, line 128:
```suggestion
			slog.Info("Failed to get item by path. It may not exist so we will create it.", "path", filename, "error", err)
```

word/pkg/graph/docs.go
```diff
@@ -17,13 +24,220 @@ func (d DocInfo) String() string {
...
+		}
+	} else {
+		doc, err = client.Drives().ByDriveId(driveID).Items().ByDriveItemId(filename).Get(ctx, nil)
+		if err != nil {
+			slog.Info("Failed to get item by ID. It may not exist so we create it.", "name", filename, "error", err)
```
g-linville, line 133:
```suggestion
			slog.Info("Failed to get item by ID. It may not exist so we will create it.", "name", filename, "error", err)
```

word/pkg/graph/docs.go
```diff
@@ -17,13 +24,220 @@ func (d DocInfo) String() string {
...
+	requestInfo.SetStreamContentAndContentType(content, "text/plain")
+
+	if doc == nil {
+		slog.Info("File does not exist. Creating.", "name", filename)
+		requestInfo.UrlTemplate = "{+baseurl}/drives/{driveid}/items/{parentid}:/{filename}:/content"
+		requestInfo.PathParametersAny = map[string]any{
+			"driveid":  driveID,
+			"parentid": parentID,
+			// URL-encode the filename if necessary.
+			"filename": filename,
```
g-linville, line 153: Doesn't look like this is getting URL-encoded?

word/pkg/graph/docs.go
```diff
@@ -17,13 +24,220 @@ func (d DocInfo) String() string {
...
 		return nil, err
 	}
 
-	docs, err := c.Drives().ByDriveId(deref(drive.GetId())).SearchWithQ(ptr("docx")).GetAsSearchWithQGetResponse(ctx, nil)
+	opts := &drives.ItemSearchWithQRequestBuilderGetRequestConfiguration{
+		QueryParameters: &drives.ItemSearchWithQRequestBuilderGetQueryParameters{
+			// Request that these fields are returned in the response.
+			Select: []string{"id", "name", "parentReference"},
+		},
+		// You can also set headers or options if needed.
```
g-linville, line 235:
```suggestion
```

word/pkg/graph/docs.go
```diff
@@ -39,6 +253,36 @@ func ListDocs(ctx context.Context, c *msgraphsdkgo.GraphServiceClient) ([]DocInf
...
+	doc, err := getItemByPath(ctx, c, deref(drive.GetId()), path)
+	if err != nil {
+		if strings.Contains(err.Error(), "item not found") {
+			return "", fmt.Errorf("doc not found")
+		}
+		return "", err
+	}
+
+	parentPath := deref(doc.GetParentReference().GetPath())
+	slog.Info("Getting doc by path", "path", path, "parentPath", parentPath)
```
g-linville, line 271: Unhelpful for the LLM

```suggestion
```

word/pkg/convert/convert.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, err
+	}
+	_ = tempfile.Close()
+	defer os.Remove(p)
+
+	var cmd *exec.Cmd
+	var outputFile string
+	if _, err := exec.LookPath("pandoc"); err == nil {
+		cmd, outputFile = pandocCmd(p)
+		slog.Info("Used pandoc to convert markdown to docx", "input", p, "output", outputFile)
```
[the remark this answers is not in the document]
iwilltry42: Yeah, slog logs to stderr by default unless configured otherwise.
I would be fine with deleting the logs or changing the level though.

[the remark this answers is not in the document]
g-linville: Oh, that's fine with me as long as it doesn't go to the LLM

word/credential/tool.gpt
```diff
@@ -16,6 +16,8 @@ Tools: ../../oauth2
 			"Fields.Read",
```
StrongMonkey, line 16: I think you should replace `Fields.Read` to `File.Read`.  I think `"Fields.Read" are added accidentlly as a typo. The scope doesn't exist

thedadams: Yes, this was my fault.

thedadams (approved):

thedadams (approved):
This looks good to me once the invalid `Fields` scopes are removed.

word/pkg/graph/docs.go
iwilltry42: Correct, that broke everything so I removed it :D

word/credential/tool.gpt
iwilltry42: I guess they could be removed as we have ReadWrite below, but I'll leave them here as the Azure permissions played tricks on me more than once :D

PR #418, opened 2025-02-06
fix: use fusejs to filter search results for slack channels and users

njhale:
Update the `Search Channels` and `Search Users` tools to use Fuse.js to filter slack channel names and user names by similarity to the query string.

Note: This may not return exact matches, but will allow LLMs to filter the channels and users down to results with similar names.

Addresses https://github.com/obot-platform/obot/issues/1660

### DIFF
--- a/slack/src/tools.js
+++ b/slack/src/tools.js
@@ -46,7 +46,8 @@ export async function searchChannels(webClient, query) {
         result?.channels ?? [],
         {
             keys: ['name'],
-            threshold: 0.4
+            threshold: 0.4,
+            findAllMatches: true
         }
     )).search(query).map(result => result.item)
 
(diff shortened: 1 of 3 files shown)


g-linville (approved):

slack/src/tools.js
```diff
@@ -11,43 +12,52 @@ export async function userContext(webClient) {
...
 }
 
 export async function searchChannels(webClient, query) {
-    const publicResult = await webClient.conversations.list({limit: 100, types: 'public_channel'})
-    const privateResult = await webClient.conversations.list({limit: 100, types: 'private_channel'})
-
-    const publicChannels = publicResult.channels.filter(channel => channel.name.includes(query))
-    const privateChannels = privateResult.channels.filter(channel => channel.name.includes(query))
+    const result = await webClient.conversations.list({limit: 100, types: 'public_channel,private_channel'})
+    const channels = (new Fuse(
```
njhale, line 45: we can improve the results for `searchUsers` as with this as well. maybe its worth doing it in this PR as well?

g-linville: sure I think that makes sense to do here

njhale: done

g-linville (approved):

cjellick (approved):

thedadams (approved):

PR #426, opened 2025-02-07
enhance: support media posts for wordpress tools

tybalex:
To address https://github.com/obot-platform/obot/issues/1682
- add tools to CRUD media files in the workspace, including image/video/audio, etc.
- other minor prompt fixes.

### DIFF
--- a/wordpress/tools/media.py
+++ b/wordpress/tools/media.py
@@ -2,7 +2,6 @@ from tools.helper import (
     WORDPRESS_API_URL,
     tool_registry,
     is_valid_iso8601,
-    str_to_bool,
     load_from_gptscript_workspace,
 )
 import os
@@ -17,7 +16,6 @@ def _format_media_response(response_json: dict):
         "id",
         "date",
         "date_gmt",
-        "guid",
         "modified",
         "modified_gmt",
         "slug",
@@ -163,8 +161,8 @@ def upload_media(client):
     try:
         data = load_from_gptscript_workspace(file_path)
     except Exception as e:
-        raise ValueError(
-            f"Failed to load file {file_path} from GPTScript workspace. Error: {e}"
+        raise Exception(
+            f"Failed to load file {file_path} from GPTScript workspace. Exception: {e}"
         )
 
     # with open(file_path, "rb") as file:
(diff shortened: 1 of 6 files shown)


wordpress/tools/media.py
```diff
@@ -0,0 +1,236 @@
...
+    context = os.getenv("CONTEXT", "view").lower()
+    context_enum = {"view", "embed", "edit"}
+    if context not in context_enum:
+        raise ValueError(
+            f"Error: Invalid context: {context}. context must be one of: {context_enum}."
+        )
+    url += f"?context={context}"
+    password = os.getenv("PASSWORD", None)
+    if password:
+        url += f"&password={password}"
```
g-linville, line 49: Should the password be URL-encoded?

wordpress/tools/media.py
```diff
@@ -0,0 +1,236 @@
...
+                f"Error: Invalid modified_before: {modified_before}. modified_before must be a valid ISO 8601 date string, in the format of YYYY-MM-DDTHH:MM:SS, or YYYY-MM-DDTHH:MM:SS+HH:MM."
+            )
+        suffix += f"&modified_before={urllib.parse.quote(modified_before)}"
+    order = os.getenv("ORDER", "desc").lower()
+    order_enum = ["asc", "desc"]
+    if order not in order_enum:
+        raise ValueError(
+            f"Error: Invalid order: {order}. order must be one of: {order_enum}."
+        )
+    suffix += f"&order={order}"
```
g-linville, line 132: I think most of this stuff getting added to the suffix should be URL-encoded for safety.

tybalex: good point I will address those

wordpress/tools/media.py
tybalex: yes. thanks for pointing it out

wordpress/tools/media.py
tybalex: updated

wordpress/tools/media.py
tybalex: updated

g-linville (approved):

PR #430, opened 2025-02-10
chore: add Google Search Console

cloudnautique:
Adds the new tool for Google Search Console
Adds new scopes to the Google Credential

### DIFF
--- a/google/credential/tool.gpt
+++ b/google/credential/tool.gpt
@@ -21,6 +21,8 @@ Tools: ../../oauth2
             "https://mail.google.com/",
             "https://www.googleapis.com/auth/gmail.readonly",
             "https://www.googleapis.com/auth/gmail.compose",
+            "https://www.googleapis.com/auth/webmasters",
+            "https://www.googleapis.com/auth/webmasters.readonly",
             "https://www.googleapis.com/auth/drive"
 		]
 	}
--- a/google/search-console/index.js
+++ b/google/search-console/index.js
@@ -7,7 +7,6 @@ const oauthToken = process.env.GOOGLE_SEARCH_OAUTH_TOKEN;
 
 async function main() {
     const command = process.argv[2];
-    console.log(command);
     if (!oauthToken) {
         console.error('Error: OAUTH_TOKEN environment variable is not set.');
         return;
--- a/google/search-console/index.js
+++ b/google/search-console/index.js
@@ -1,0 +1,99 @@
+import { fetchGSCProperties } from './src/siteList.js';
+import { getUrlsForProperty } from './src/urlsForProperty.js';
+import { inspectUrl } from './src/urlInspection.js';
+import { getTopKeywordsForUrl } from './src/keywords.js';
+
+const oauthToken = process.env.GOOGLE_SEARCH_OAUTH_TOKEN;
+
+async function main() {
+    const command = process.argv[2];
+    console.log(command);
+    if (!oauthToken) {
+        console.error('Error: OAUTH_TOKEN environment variable is not set.');
+        return;
+    }
+
+    const propertyUrl = process.env.PROPERTY;
+    const url = process.env.URL;
+
+    switch (command) {
+        case 'getUrlsForProperty':
+            if (!propertyUrl) {
+                console.error('Error: PROPERTY environment variable is not set.');
+                return;
+            }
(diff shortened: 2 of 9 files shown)


google/credential/tool.gpt
```diff
@@ -21,6 +21,8 @@ Tools: ../../oauth2
             "https://mail.google.com/",
             "https://www.googleapis.com/auth/gmail.readonly",
             "https://www.googleapis.com/auth/gmail.compose",
+            "https://www.googleapis.com/auth/webmasters",
+            "https://www.googleapis.com/auth/webmasters.readonly",
```
thedadams, line 25: Does adding these scopes break existing credentials?

google/search-console/index.js
```diff
@@ -0,0 +1,99 @@
+import { fetchGSCProperties } from './src/siteList.js';
+import { getUrlsForProperty } from './src/urlsForProperty.js';
+import { inspectUrl } from './src/urlInspection.js';
+import { getTopKeywordsForUrl } from './src/keywords.js';
+
+const oauthToken = process.env.GOOGLE_SEARCH_OAUTH_TOKEN;
+
+async function main() {
+    const command = process.argv[2];
+    console.log(command);
```
thedadams, line 10:
```suggestion
```

google/credential/tool.gpt
njhale: IIUC, they'll at least need to be added to the [UI's instructions for configuring the oauth app](https://github.com/obot-platform/obot/blob/4e0b2562481354097066610b8e521b84f7a61f13/ui/admin/app/lib/model/oauthApps/providers/google.ts#L16)

google/search-console/index.js
```diff
@@ -0,0 +1,98 @@
...
+import { getUrlsForProperty } from './src/urlsForProperty.js';
+import { inspectUrl } from './src/urlInspection.js';
+import { getTopKeywordsForUrl } from './src/keywords.js';
+
+const oauthToken = process.env.GOOGLE_SEARCH_OAUTH_TOKEN;
+
+async function main() {
+    const command = process.argv[2];
+    if (!oauthToken) {
+        console.error('Error: OAUTH_TOKEN environment variable is not set.');
```
njhale, line 11: nit:

```suggestion
        console.error('Error: GOOGLE_SEARCH_OAUTH_TOKEN environment variable is not set.');
```

google/credential/tool.gpt
cloudnautique: If my understanding is correct, if an existing credential was authenticated for a tool that was updated to require a new scope would see an unauthorized response. They would need to delete the credential and reauthenticate. 

In this case, we are adding an entirely new tool / bundle so it would only be impactful if it was added to an existing agent and picked up in the same thread.

google/search-console/index.js
```diff
@@ -0,0 +1,98 @@
...
+        return;
+    }
+
+    const propertyUrl = process.env.PROPERTY;
+    const url = process.env.URL;
+
+    switch (command) {
+        case 'getUrlsForProperty':
+            if (!propertyUrl) {
+                console.error('Error: PROPERTY environment variable is not set.');
```
njhale, line 21: These are fatal errors, so I think we'll want to exit with a non-zero exit code so GPTScript returns the error to the LLM.

google/search-console/src/keywords.js
```diff
@@ -0,0 +1,79 @@
...
+function getToday() {
+    const today = new Date();
+    return today.toISOString().split("T")[0];
+}
+
+function getThreeMonthsAgo() {
+    const date = new Date();
+    date.setMonth(date.getMonth() - 3);
+    return date.toISOString().split("T")[0];
+}
```
njhale, line 79: I think the relative dates should be adjusted by the user's timezone. It's injected into the tools environment as `OBOT_USER_TIMEZONE`.

njhale (changes requested):

google/search-console/index.js
cloudnautique: Done.

google/search-console/index.js
cloudnautique: Good catch! updated to use the variable set by the google cred.

njhale (approved):

drpebcak (approved):

PR #434, opened 2025-02-12
feat: safesearch variant of the tavily search to that only allows searching a limited set of domains

iwilltry42:
https://github.com/obot-platform/obot/issues/1700

The idea is that the UI will set the TAVILY_ALLOWED_DOMAINS* env vars such that the LLM cannot decide the search domains itself.

### DIFF
--- a/search/tavily/main.py
+++ b/search/tavily/main.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import sys
+from typing import List
 
 from tavily import TavilyClient
 from urllib3.util import parse_url
@@ -15,7 +16,7 @@ def main():
     client = TavilyClient()  # env TAVILY_API_KEY required
 
     match command:
-        case "search":
+        case "search" | "safe-search":
             query = os.getenv("QUERY", "").strip()
             if not query:
                 print("No search query provided")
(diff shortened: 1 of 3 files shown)


search/tavily/main.py
```diff
@@ -83,5 +75,44 @@ def main():
...
+
+    if len(allowed_domains) == 0:
+        print("No allowed domains provided")
+        sys.exit(1)
+
+    # allow not setting INCLUDE_DOMAINS -  fallback to all allowed domains
+    if len(include_domains) == 0:
+        include_domains = allowed_domains
+
+    if len(allowed_domains) > 0:
```
g-linville, line 94: This is a redundant `if` statement, since we already exit if `len(allowed_domains) == 0` earlier in this function.

search/tavily/main.py
```diff
@@ -83,5 +75,44 @@ def main():
...
+        domain.strip() for domain in allowed_domains_str.split(",") if domain.strip()
+    ]
+
+    if len(allowed_domains) == 0:
+        print("No allowed domains provided")
+        sys.exit(1)
+
+    # allow not setting INCLUDE_DOMAINS -  fallback to all allowed domains
+    if len(include_domains) == 0:
+        include_domains = allowed_domains
```
g-linville, line 92: I think we can just return `allowed_domains` right here instead of proceeding with the rest of the function.

g-linville (changes requested):

search/tavily/safesearch.gpt
```diff
@@ -0,0 +1,37 @@
...
+#!/bin/bash
+
+if [ -z "${TAVILY_ALLOWED_DOMAINS}" ]; then
+  echo "No allowed domains provided - do not use the tavily-safe-search tool"
+  exit 0
+else
+
+  if [ -z "${TAVILY_ALLOWED_DOMAINS_DESCRIPTION}" ]; then
+    TAVILY_ALLOWED_DOMAINS_DESCRIPTION="No description provided, but you have access to search across the following domains: ${TAVILY_ALLOWED_DOMAINS}"
+  fi
```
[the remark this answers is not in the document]
iwilltry42: It's actually just setting this variable to use it further down :+1:

search/tavily/main.py
iwilltry42: Yup.. refactored without looking around

search/tavily/main.py
iwilltry42: absolutely

search/tavily/safesearch.gpt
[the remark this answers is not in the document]
g-linville: My bad lol

g-linville (approved):

PR #436, opened 2025-02-12
feat: add slack daemon trigger provider

njhale:
Add daemon trigger provider for slack.

Addresses https://github.com/obot-platform/obot/issues/1404


njhale:
Closing this out. We're going with another design for supporting slack triggers.

PR #437, opened 2025-02-12
enhance: use workspace file revisions to avoid clobbering the database

njhale:
Addresses https://github.com/obot-platform/obot/issues/1813


database/main.go
```diff
@@ -108,7 +127,9 @@ func saveWorkspaceDB(
 	}
 
 	if err := g.WriteFileInWorkspace(ctx, dbWorkspacePath, updatedDBData, gptscript.WriteFileInWorkspaceOptions{
-		WorkspaceID: workspaceID,
+		WorkspaceID:    workspaceID,
+		CreateRevision: &([]bool{true}[0]),
+		LatestRevision: revisionID,
 	}); err != nil {
 		return fmt.Errorf("Error writing updated DB file to workspace: %v", err)
 	}
```
njhale, line 135: Not sure I'm using the feature correctly, but if this is right I'll need to delete the last revision after creating the new one.

database/go.mod
```diff
@@ -2,21 +2,23 @@ module obot-platform/database
 
 go 1.23.3
 
+replace github.com/gptscript-ai/go-gptscript => github.com/thedadams/go-gptscript v0.0.0-20250219113618-25d959a071ff
```
njhale, line 5: I need https://github.com/gptscript-ai/go-gptscript/pull/93 to merge before we can remove this and merge this PR.

database/main.go
```diff
@@ -108,11 +120,38 @@ func saveWorkspaceDB(
...
+	})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error listing revisions: %v\n", err)
+		return nil
+	}
+
+	lastRevisionIndex := slices.IndexFunc(revisions, func(rev gptscript.FileInfo) bool {
+		return rev.RevisionID == revisionID
+	})
+
```
thedadams, line 142: Revisions should come out in order.

database/main.go
```diff
@@ -108,11 +120,38 @@ func saveWorkspaceDB(
...
+	if lastRevisionIndex < 0 {
+		return nil
+	}
+
+	for _, rev := range revisions[:lastRevisionIndex+1] {
```
thedadams, line 147: This seems to be delete ALL revisions. Is that the intention?

workspace-files/main.go
```diff
@@ -174,17 +174,17 @@ func read(ctx context.Context, filename string) error {
 		return err
 	}
 
-	data, err := client.ReadFileInWorkspace(ctx, path.Join(FilesDir, filename))
+	response, err := client.ReadFileInWorkspace(ctx, path.Join(FilesDir, filename))
```
thedadams, line 177: In the latest version of the go-gptscript changes, this function returns `[]byte` instead of the data structure.

njhale: Ah, meant to drop this. Thanks for catching.

database/main.go
njhale: Iiuc, this drops all revisions up-to and including the previous revision (not the one we just created).

thedadams (approved):

database/main.go
```diff
@@ -108,11 +120,38 @@ func saveWorkspaceDB(
...
 	}); err != nil {
 		return fmt.Errorf("Error writing updated DB file to workspace: %v", err)
 	}
 
+	// Delete old revisions after successfully writing the new revision
+	revisions, err := g.ListRevisionsForFileInWorkspace(ctx, dbWorkspacePath, gptscript.ListRevisionsForFileInWorkspaceOptions{
+		WorkspaceID: workspaceID,
+	})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error listing revisions: %v\n", err)
```
StrongMonkey, line 135: Do we want to intentionally ignore error here?

database/main.go
```diff
@@ -108,11 +120,38 @@ func saveWorkspaceDB(
...
+
+	if lastRevisionIndex < 0 {
+		return nil
+	}
+
+	for _, rev := range revisions[:lastRevisionIndex+1] {
+		if err := g.DeleteRevisionForFileInWorkspace(ctx, dbWorkspacePath, rev.RevisionID, gptscript.DeleteRevisionForFileInWorkspaceOptions{
+			WorkspaceID: workspaceID,
+		}); err != nil {
+			fmt.Fprintf(os.Stderr, "Error deleting revision %s: %v\n", rev.RevisionID, err)
```
StrongMonkey, line 151: Same here

database/main.go
njhale: Yes, I don't think deleting old revisions should be a critical error. IIUC, this will return the error to Obot, but not the LLM. Correct me if I'm wrong there though.

StrongMonkey (approved):

PR #440, opened 2025-02-13
fix: github: return user id

g-linville:
This is needed for https://github.com/obot-platform/obot/issues/1747

### DIFF
--- a/auth-providers-common/pkg/state/state.go
+++ b/auth-providers-common/pkg/state/state.go
@@ -40,33 +40,10 @@ func ObotGetState(p *oauth2proxy.OAuthProxy) http.HandlerFunc {
 
 		reqObj.Header = sr.Header
 
-		state, err := p.LoadCookiedSession(reqObj)
+		ss, err := GetSerializableState(p, reqObj)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


auth-providers-common/pkg/state/state.go
```diff
@@ -40,40 +40,45 @@ func ObotGetState(p *oauth2proxy.OAuthProxy) http.HandlerFunc {
...
 
-		var setCookies []string
-		if state.IsExpired() || (p.CookieOptions.Refresh != 0 && state.Age() > p.CookieOptions.Refresh) {
-			setCookies, err = refreshToken(p, reqObj)
-			if err != nil {
-				http.Error(w, fmt.Sprintf("failed to refresh token: %v", err), http.StatusForbidden)
-				return
-			}
-		}
+func GetSerializableState(p *oauth2proxy.OAuthProxy, r *http.Request) (SerializableState, error) {
```
g-linville, line 56: This is a refactoring of some of the `ObotGetState` behavior so that I could call just this function to get the SerializableState, and then modify it afterwards. We needed this for GitHub.

thedadams (approved):

drpebcak (approved):

PR #450, opened 2025-02-19
enhance: improve WordPress Tool prompts 

tybalex:
- prompt the wordpress tool so it reminds the user that draft post is not visible to user without permissions.

### DIFF
--- a/wordpress/tool.gpt
+++ b/wordpress/tool.gpt
@@ -194,6 +194,7 @@ WordPress posts support the following formats:
 - chat: A chat transcript.
 
 Before using the Create Post Tool, you MUST confirm the post title and content with the user.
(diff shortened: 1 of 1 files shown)


g-linville (approved):

thedadams:
As far as I can tell, this doesn't "fix" the issue, per se. It introduces a warning message from the LLM.

Is there a way to change the URL so the user will be redirected to a login page if they aren't authenticated?

tybalex:
> Is there a way to change the URL so the user will be redirected to a login page if they aren't authenticated?

This is a good idea, However, technically we have no idea if the users have logged-in to the site in their browsers or not.
So I guess this change would need to be made on the WordPress site's end:
WordPress Site administrators can achieve this by editing the relevant PHP files or by installing a plugin that redirects draft post pages for users who are not logged in.

PR #451, opened 2025-02-19
Enhance: Add support to pull slack bot token

StrongMonkey:
Add support to use slack bot token to send message a bot. This is done by adding a new tool into slack bundle that consumes a separate credential that requests bot token.

### DIFF
--- a/slack/credential/tool.gpt
+++ b/slack/credential/tool.gpt
@@ -12,7 +12,7 @@ Tools: ../../oauth2
 	"oauthInfo": {
 		"integration": "slack",
 		"token": "SLACK_TOKEN",
-		"scope": [
+		"userScope": [
 			"channels:history",
 			"groups:history",
 			"im:history",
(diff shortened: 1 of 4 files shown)


slack/credential/tool.gpt
```diff
@@ -12,7 +12,7 @@ Tools: ../../oauth2
 	"oauthInfo": {
 		"integration": "slack",
 		"token": "SLACK_TOKEN",
-		"scope": [
+		"userScope": [
```
StrongMonkey, line 15: I am changing this so that it is consistent with slack.

g-linville (approved):
This LGTM, but I think we should consider whether we want to split bot tools out into their own separate bundle. Also I think it would be good to add a tool to send messages to channels as a bot, rather than only DMs.

StrongMonkey:
@g-linville 
>This LGTM, but I think we should consider whether we want to split bot tools out into their own separate bundle. Also I think it would be good to add a tool to send messages to channels as a bot, rather than only DMs.

Yeah, I was thinking about pulling it as a separate bundle/tool too, wasn't super comfortable right now to have a separate slack tool to do thing as bot. But yeah we could have the discussion.

thedadams (approved):

StrongMonkey:
@g-linville I am merging this as a tool in existing bundle for now. Will consider adding message to channel later, this is more completing the demo use-case for unsubscribe agent.

PR #454, opened 2025-02-20
Enhance: Add slack bot tool as a bundle

StrongMonkey:
Make slack bot tools as a separate bundle tool by consuming bot token credential. This is so that we can avoid the case where we add more bot tools into our existing slack tool cause user to have prompt twice for different credentials requesting user vs bot token.

To minimize the scope of this PR, it is only adding functionality to send DM as Bot. We can extend more tools based on real use case later, like the tools we already have for slack

### DIFF
--- a/slack/tool-bot.gpt
+++ b/slack/tool-bot.gpt
@@ -1,0 +1,44 @@
+---
+Name: Slack Bot
+Description: Tools for interacting with Slack as an App
+Metadata: bundle: true
+Share Tools: Send Direct Message As Bot
+
+---
+Name: Send Direct Message As Bot
+Description: Send a direct message as a bot or an app in the Slack workspace
+Credential: ./credential/bot.gpt
+Share Tools: List Users As Bot, Search Users As Bot
+Param: userids: comma-separated list of user IDs to send the message to for a group message (example: USER1ID,USER2ID), or just one ID for an individual message
+Param: text: the text to send
+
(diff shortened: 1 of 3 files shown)


thedadams (approved):

slack/tool-bot.gpt
```diff
@@ -0,0 +1,44 @@
+---
+Name: Slack Bot
+Description: Tools for interacting with Slack as an App
+Metadata: bundle: true
+Share Tools: Send Direct Message As Bot
```
g-linville, line 5:
```suggestion
Share Tools: Send Direct Message As Bot, List Users As Bot, Search Users As Bot
```

g-linville (changes requested):

slack/tool-bot.gpt
StrongMonkey: Intentionally avoid adding these two tools into share tools so that these are not indexed in obot. For now I only want to expose `Send Direct Message As Bot` since that's the only use case we have as a bot.

g-linville: Oh sorry, I missed the comment you made about this in the PR description

g-linville (approved):

slack/tool-bot.gpt
StrongMonkey: No problem 👍

PR #457, opened 2025-02-21
Fix: list/search channel properly using cursor

StrongMonkey:
Fix the channel list by using cursor to find the rest of the results, also use formatter to format code properly.

https://github.com/obot-platform/obot/issues/1827

### DIFF
--- a/slack/src/tools.js
+++ b/slack/src/tools.js
@@ -3,523 +3,569 @@ import { Mutex } from "async-mutex"
 import Fuse from "fuse.js"
 
 export async function userContext(webClient) {
-    const result = await webClient.auth.test({})
-    const userResult = await webClient.users.info({user: result.user_id})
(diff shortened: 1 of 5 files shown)


slack/src/tools.js
```diff
@@ -3,523 +3,569 @@ import { Mutex } from "async-mutex"
...
-            }
-        })
-        const datasetID = await gptscriptClient.addDatasetElements(elements, {
-            name: 'slack_channels',
-        })
-        console.log(`Created dataset with ID ${datasetID} with ${channels.length} channels`)
-    } catch (e) {
-        console.log('Failed to create dataset:', e)
-    }
+  let allChannels = []
```
StrongMonkey, line 15: @njhale a lot of formatting changes, but this is the change that i made so that it will follow the cursor and list the rest of the channels. I feel like it is being paginated on their api so we have to go through all the pages.

PR #464, opened 2025-03-04
feat: reuse documents/embeddings for files with the same sha256sum (+fix pgx bug)

iwilltry42:
- calculate sha256sum for the content of each ingested file and record it in the DB
- when ingesting a file, check if there's a file with the same checksum in the DB
- if yes and the embeddingModel (config on the owning dataset) matches the incoming one, then fetch the file's documents and their embeddings to re-use for the incoming file

NOTE: This is fine in the Obot setup, as we're using a static flow config for Obot, but if someone is playing with the flow config to test different settings, this may lead to reusing documents from file that were ingested with an old config - we can add checks for that as a follow-up.

Also, I figured that there's a race condition in the pgvector implementation, because appending to the pgx batch queue is not lock protected and I saw some cases, where documents got lost due to this. This is fixed now.

Issue: https://github.com/obot-platform/obot/issues/1803

### DIFF
--- a/knowledge/pkg/vectorstore/pgvector/pgvector.go
+++ b/knowledge/pkg/vectorstore/pgvector/pgvector.go
@@ -354,7 +354,6 @@ func (v VectorStore) AddDocuments(ctx context.Context, docs []vs.Document, colle
 	slog.Debug("Sending batch to pgvector", "store", "pgvector", "batchSize", b.Len())
 
 	results := v.conn.SendBatch(ctx, b)
-	defer results.Close()
 	for _, d := range docs {
 		_, err := results.Exec()
 		if err != nil {
--- a/knowledge/pkg/vectorstore/pgvector/pgvector.go
+++ b/knowledge/pkg/vectorstore/pgvector/pgvector.go
@@ -307,13 +307,12 @@ func (v VectorStore) AddDocuments(ctx context.Context, docs []vs.Document, colle
 		VALUES($1, $2, $3, $4, $5)`, v.embeddingTableName)
 
 	var wg sync.WaitGroup
+	qqLock := sync.Mutex{} // lock for the pgx batch queue
 	semaphore := make(chan struct{}, v.embeddingConcurrency)
+	wg.Add(len(docs))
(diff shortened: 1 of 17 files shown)


knowledge/pkg/vectorstore/pgvector/pgvector.go
```diff
@@ -349,7 +351,18 @@ func (v VectorStore) AddDocuments(ctx context.Context, docs []vs.Document, colle
 		return nil, sharedErr
 	}
 
-	return ids, v.conn.SendBatch(ctx, b).Close()
+	slog.Debug("Sending batch to pgvector", "store", "pgvector", "batchSize", b.Len())
+
+	results := v.conn.SendBatch(ctx, b)
+	defer results.Close()
```
g-linville, line 357: I don't think we want to defer this, since we call it on the same line as `return` later, and there are no other return statements prior to then.

knowledge/pkg/vectorstore/sqlite-vec/sqlite-vec.go
```diff
@@ -317,6 +317,45 @@ func (v *VectorStore) RemoveDocument(ctx context.Context, documentID string, col
...
+
+	doc.ID = documentID
+
+	// Get embeddings
+	var emb []byte
+	err = v.db.Raw(fmt.Sprintf(`
+		SELECT embedding
+		FROM [%s_vec]
+		WHERE document_id = ?;
+		`), documentID).Row().Scan(&emb)
```
g-linville, line 345: This looks like an incomplete `Sprintf` since no value is provided for the `%s` that's in there.

g-linville (changes requested):

knowledge/pkg/vectorstore/pgvector/pgvector.go
iwilltry42: True, I removed that other return.. Close() is idempotent here, so it didn't do any harm :sweat_smile:

g-linville (approved):

thedadams (approved):

PR #465, opened 2025-03-04
enhance: add support to get past meeting instances

tybalex:
https://github.com/obot-platform/obot/issues/1892

This PR:
- add API to get past meeting instances of a meeting. The information can then be used to get recordings/summaries ..etc
- tweak prompt so agent has aware of date & time
- allow auto-start-ai-summary for meeting creation
- add instructions about instant meetings and personal meeting rooms
- add logger and other minor fix.


zoom/tool.gpt
```diff
@@ -133,19 +135,27 @@ Share Context: Zoom Context
...
 Credential: ./credential
 Share Context: Zoom Context
-Param: meeting_id: The ID of the meeting to get the details for
+Param: meeting_id_or_uuid: The ID or UUID of the meeting to get the details for. If providing the meeting ID instead of UUID, the response will be for the latest meeting instance.
 
 #!/usr/bin/env python3 ${GPTSCRIPT_TOOL_DIR}/main.py GetPastMeetingDetails
 
+---
+Name: List Past Meeting Instances
+Description: List all instances of a Zoom meeting.
```
g-linville, line 147: Same here.

zoom/tools/helper.py
```diff
@@ -8,6 +10,37 @@
...
+    # Create a logger
+    logger = logging.getLogger(name)
+    logger.setLevel(logging.DEBUG)  # Set the logging level
+
+    # Create a stream handler that writes to sys.stderr
+    stderr_handler = logging.StreamHandler(sys.stderr)
+
+    # Create a log formatter
+    formatter = logging.Formatter(
+        "[WordPress Tool Debugging Log]: %(asctime)s - %(name)s - %(levelname)s - %(message)s"
```
g-linville, line 31:
```suggestion
        "[Zoom Tool Debugging Log]: %(asctime)s - %(name)s - %(levelname)s - %(message)s"
```

zoom/tools/meetings.py
```diff
@@ -416,6 +427,42 @@ def update_meeting():
...
+        settings["auto_recording"] = audio_recording
+    if "CONTACT_EMAIL" in os.environ:
+        contact_email = os.environ["CONTACT_EMAIL"]
+        settings["contact_email"] = contact_email
+    if "CONTACT_NAME" in os.environ:
+        contact_name = os.environ["CONTACT_NAME"]
+        settings["contact_name"] = contact_name
+    if "PRIVATE_MEETING" in os.environ:
+        private_meeting = str_to_bool(os.environ["PRIVATE_MEETING"])
+        settings["private_meeting"] = private_meeting
```
g-linville, line 443: nit: any reason not to do this?

```suggestion
    if "AUDIO_RECORDING" in os.environ:
        settings["auto_recording"] = os.environ["AUDIO_RECORDING"]
    if "CONTACT_EMAIL" in os.environ:
        settings["contact_email"] = os.environ["CONTACT_EMAIL"]
    if "CONTACT_NAME" in os.environ:
        settings["contact_name"] = os.environ["CONTACT_NAME"]
    if "PRIVATE_MEETING" in os.environ:
        settings["private_meeting"] = str_to_bool(os.environ["PRIVATE_MEETING"])
```

Also, is the param supposed to be called `auto_recording`, or `audio_recording`? I'm not sure if the inconsistency is in the name of the env var or of the parameter, but they should be changed to match.

zoom/tools/meetings.py
```diff
@@ -416,6 +427,42 @@ def update_meeting():
...
+        meeting_invitees_list = [
+            {"email": invitee} for invitee in meeting_invitees.split(",")
+        ]
+        settings["meeting_invitees"] = meeting_invitees_list
+
+    if "AUTO_START_MEETING_SUMMARY" in os.environ:
+        auto_start_meeting_summary = str_to_bool(
+            os.environ["AUTO_START_MEETING_SUMMARY"]
+        )
+        settings["auto_start_meeting_summary"] = auto_start_meeting_summary
```
g-linville, line 464:
```suggestion
    if "AUTO_START_MEETING_SUMMARY" in os.environ:
        settings["auto_start_meeting_summary"] = str_to_bool(os.environ["AUTO_START_MEETING_SUMMARY"])
```

zoom/tools/helper.py
tybalex: Thanks for the catch lol

zoom/tools/meetings.py
tybalex: > nit: any reason not to do this?
> 
> Also, is the param supposed to be called `auto_recording`, or `audio_recording`? I'm not sure if the inconsistency is in the name of the env var or of the parameter, but they should be changed to match.

and yes, we should simplify these. it was there probably because some of the env needs condition checks.
Fixed as requested.

zoom/tool.gpt
```diff
@@ -133,19 +135,27 @@ Share Context: Zoom Context
 
 ---
 Name: Get Past Meeting Details
-Description: Get the details of a past Zoom meeting.
+Description: Get the details of a Zoom meeting instance. It is possible that the meeting has not occurred yet.
```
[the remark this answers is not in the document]
tybalex: Good question! the name comes from Zoom's API. 
It can only return the `details` of a past meeting instance, however, it is possible to pass-in a future meeting instance UUID, though it will only return something like `the meeting instance doesn't exist yet`.
So I guess it's ok to keep the origin name.

zoom/tool.gpt
tybalex: same as above, `instances` indicates it is likely a past meeting instance

zoom/tool.gpt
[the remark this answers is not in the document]
g-linville: Oh, yeah let's just keep the current name.

PR #467, opened 2025-03-04
enhance: github: add tool to get job logs

g-linville:
This will be needed for the GitHub CI failure webhook obot.

### DIFF
--- a/github/index.js
+++ b/github/index.js
@@ -20,6 +20,7 @@ import {
     addIssueLabels,
     removeIssueLabels,
     getUser,
+    getJobLogs,
 } from './src/tools.js';
 
 if (process.argv.length !== 3) {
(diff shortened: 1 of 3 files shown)


drpebcak (approved):

tybalex (approved):

tybalex:
Look good. One common concern with logs (not sure if it's the case for this tool): What if there are so many logs that they exceed or flood the context window?

g-linville:
Good point, @tybalex. I think I will modify it so that it takes an optional `tail` parameter that only gets the last N lines. That way if the LLM blows out its context window, it can ask again for fewer lines. Wherever the error is is likely to be near the end anyway.

g-linville:
Actually, I'm just gonna merge this as is so that I can get my obot up, and I'll add that as a follow-up.

PR #468, opened 2025-03-04
Enhance: Add more slack bot tools

StrongMonkey:
This PR adds more slack bot tools so that the bot can do more things like send message to channels, list channels. It will also request more scopes.

Search message from slack tool is not added because it is not supported by the bot.

### DIFF
--- a/slack/credential/bot.gpt
+++ b/slack/credential/bot.gpt
@@ -13,9 +13,20 @@ Tools: ../../oauth2
 		"integration": "slack",
 		"token": "SLACK_TOKEN",
 		"scope": [
+			"channels:history",
+			"groups:history",
+			"im:history",
+			"mpim:history",
+			"channels:read",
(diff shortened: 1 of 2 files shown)


StrongMonkey:
Actually, I am going to see if I can reuse some context from tool.gpt.

cjellick:
Im not sure I love this approach. Particularly having a "copy and paste" of tools from tool.gpt with just "as bot" appended to their names. Why do you do that? Is it so that both tools can co-exist without conflicting?

StrongMonkey:
@cjellick Yes, I am trying to avoid conflicts. I haven't really proved that this will cause a problem, but I have checked the llm response and toolName are just set with the normalized version of their name in gptscript. So adding the bot is the safest bet.

thedadams (approved):

g-linville (approved):

PR #472, opened 2025-03-06
chore: weather tool

drpebcak:


thedadams (approved):
This looks good, but should we also have a tool that gets the coordinates for a given city and state?

drpebcak:
> This looks good, but should we also have a tool that gets the coordinates for a given city and state?

gpt-4o is pretty good at knowing the cords already, so it shouldnt be an issue. If we did have a tool for doing this, i think it would just be a canned google search essentially.

PR #474, opened 2025-03-07
fix: bsky create post link and tag formatting

njhale:
Links and tags in posts created by the `Bluesky` bundle's `Create Post`
tool are not being rendered correctly on bsky.app. This is because the
posts were missing metadata that Bluesky requires to render them.

To correct this, resolve and include the missing metadata with posts.

e.g. Before this change

Screenshot 2025-03-06 at 8 54 17 PM

e.g. After

Screenshot 2025-03-06 at 9 05 30 PM

Addresses https://github.com/obot-platform/obot/issues/1960

### DIFF
--- a/bluesky/src/embed.ts
+++ b/bluesky/src/embed.ts
@@ -1,0 +1,63 @@
+import { AtpAgent, RichText } from "@atproto/api"
+import * as cheerio from "cheerio"
+
+type Metadata = {
+  title: string
+  description: string
+  image: string
+}
+
+/**
+ * Extract Open Graph (OG) metadata from the content of page.
+ * @param url - The URL to extract metadata from
+ * @returns Metadata containing title, description, and image
+ */
+export const getPageMetadata = async (url: string): Promise<Metadata> => {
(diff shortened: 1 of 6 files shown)


thedadams (approved):

bluesky/src/embed.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (!url) return
+
+  try {
+    const metadata = await getUrlMetadata(url)
+    let thumbBlob
+
+    if (metadata.image) {
+      const blob = await fetch(metadata.image).then(r => r.blob())
+      const { data } = await agent.uploadBlob(blob, { encoding: "image/jpeg" })
```
tybalex, line 47: is fetched image always `jpeg`? Should we detect the MIME type instead

njhale: Nice catch! I change it to just passed the mime type from the blob through. Presumably, the SDKs uploadBlob function will fail if the mime-type isn't supported by the AT Proto Lexicon.

tybalex (approved):
lgtm

PR #481, opened 2025-03-08
Firecrawl Tool

sanjay920:
This PR introduces Firecrawl. In the future this tool should be able to support a locally running or self hosted firecrawl, but this starts with using Firecrawl via API token.

Get an API key from: https://www.firecrawl.dev/app


firecrawl/cmd/scrape.go
```diff
@@ -0,0 +1,44 @@
...
+
+func Scrape(ctx context.Context, apiKey, url string) (string, error) {
+	apiKey = strings.TrimSpace(apiKey)
+	if apiKey == "" {
+		return "", fmt.Errorf("API key is required")
+	}
+
+	if url == "" {
+		return "", fmt.Errorf("URL is required")
+	}
```
thedadams, line 20: Can this validation be moved to `main.go`? It seems like we would have to do this for each command we add, so it would be better suited in a more centralized location.

thedadams (changes requested):

thedadams (approved):

iwilltry42 (approved):

PR #482, opened 2025-03-10
fix: generated image download path

njhale:
Workspace files stored in the `files` directory are now served at `/api/threads/{threadId}/file/{file_name}` (`/file/` singular) instead of `/api/threads/{threadId}/files/{file_name}` (`/files/` plural). Update the `Generate Images` and
`Analyze Images` tools to comply with this change.

This fixes broken generated image links in the UI and when analyzing
them.

Addresses https://github.com/obot-platform/obot/issues/2005

### DIFF
--- a/images/src/generate.ts
+++ b/images/src/generate.ts
@@ -9,7 +9,7 @@ type ImageQuality = 'standard' | 'hd';
 
 const threadId = process.env.OBOT_THREAD_ID
 const obotServerUrl = process.env.OBOT_SERVER_URL
-const downloadBaseUrl = (threadId && obotServerUrl) ? `${obotServerUrl}/api/threads/${threadId}/files` : null
+const downloadBaseUrl = (threadId && obotServerUrl) ? `${obotServerUrl}/api/threads/${threadId}/file` : null
 
 export async function generateImages(
   prompt: string = '',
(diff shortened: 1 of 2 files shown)


iwilltry42 (approved):

images/src/generate.ts
```diff
@@ -81,14 +81,13 @@ async function download(client: gptscript.GPTScript, imageUrl: string): Promise<
...
-
   // Convert the image to webp format
-  content = await sharp(content).webp({ quality: 100 }).toBuffer()
+  content = Buffer.from(await sharp(content).webp({ quality: 100 }).toBuffer())
 
   // Generate a SHA-256 hash of the imageURL to use as the filename
   const filePath = `generated_image_${createHash('sha256').update(imageUrl).digest('hex').substring(0, 8)}.webp`;
 
-  await client.writeFileInWorkspace(`${threadId ? 'files/' : ''}${filePath}`, content);
+  await client.writeFileInWorkspace(`${threadId ? 'files/' : ''}${filePath}`, content.buffer);
```
g-linville, line 90: Should this still be `files/` or should it also be changed to `file/`?

g-linville (approved):

images/src/generate.ts
njhale: No, this shouldn't change

PR #487, opened 2025-03-11
fix: update workspace read/write method, so it's clear it only supports plain text format

tybalex:
https://github.com/obot-platform/obot/issues/1986

### DIFF
--- a/workspace-files/main.go
+++ b/workspace-files/main.go
@@ -23,6 +23,10 @@ var (
 	MaxFileSize = 250_000
 )
 
+var unsupportedFileTypes = []string{
+	".pdf", ".docx", ".doc", ".pptx", ".ppt", ".xlsx", ".xls", ".jpg", ".png", ".gif", ".mp3", ".mp4", ".zip", ".rar",
+}
+
 func main() {
 	if len(os.Args) == 1 {
 		fmt.Printf(`
@@ -169,6 +173,13 @@ func list(ctx context.Context, filename string) error {
 }
 
 func read(ctx context.Context, filename string) error {
+
+	for _, ext := range unsupportedFileTypes {
+		if strings.HasSuffix(strings.ToLower(filename), ext) {
+			return fmt.Errorf("reading files with extension %s is not supported", ext)
+		}
+	}
+
 	client, err := gptscript.NewGPTScript()
 	if err != nil {
 		return err
@@ -192,6 +203,13 @@ func read(ctx context.Context, filename string) error {
 }
 
 func write(ctx context.Context, filename, content string) error {
+	// Check if the file extension is unsupported
+	for _, ext := range unsupportedFileTypes {
+		if strings.HasSuffix(strings.ToLower(filename), ext) {
+			return fmt.Errorf("writing to files with extension %s is not supported", ext)
+		}
+	}
+
 	client, err := gptscript.NewGPTScript()
 	if err != nil {
 		return err
--- a/workspace-files/main.go
+++ b/workspace-files/main.go
@@ -174,6 +174,7 @@ func list(ctx context.Context, filename string) error {
 
 func read(ctx context.Context, filename string) error {
 
+	// Check for common binary file types that are not supported. Less common types should be caught by the utf8.Valid() check
 	for _, ext := range unsupportedFileTypes {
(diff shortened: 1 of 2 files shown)


workspace-files/main.go
```diff
@@ -169,6 +173,13 @@ func list(ctx context.Context, filename string) error {
 }
 
 func read(ctx context.Context, filename string) error {
+
+	for _, ext := range unsupportedFileTypes {
```
cjellick, line 177: i think this line below
```
	if utf8.Valid(data) {
```
handles the prevention of reading binary files. do we need both?

workspace-files/main.go
```diff
@@ -192,6 +203,13 @@ func read(ctx context.Context, filename string) error {
 }
 
 func write(ctx context.Context, filename, content string) error {
+	// Check if the file extension is unsupported
+	for _, ext := range unsupportedFileTypes {
```
cjellick, line 207: i dont love having a denylist here has it can be infinitely long, but i guess this is a reasonable "most common" list. i think its probably ok

workspace-files/tool.gpt
```diff
@@ -38,14 +38,14 @@ Params: dir: The directory to list files from
 
 ---
 Name: workspace_read
-Description: Read the contents of a file in the workspace
+Description: Read the contents of a file in the workspace. Supports only plain text formats (e.g., .txt, .md) and does not allow writing to binary files (e.g., .pdf, .docx, .pptx, .xlsx, .jpg, .png, .gif, .mp3, .mp4, .zip, .rar).
```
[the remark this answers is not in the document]
cjellick: honestly, logically it feels like just `Supports only plain text formats (e.g., .txt, .md)` should be enough. WDYT?

workspace-files/main.go
tybalex: `utf8.Valid(data)` is more like validating  whether the data slice contains only valid UTF-8 encoded characters. it can also handle the case that the input text file is somehow corrupted.

On the other hand, this list of common unsupported binary files is used to filter out unsupported files. 

I decided to go with `unsupported` list because technically  text input file with any non-binary extension (such as .log, .mdtxt, .ipynb, etc) should be allowed, so I guess it's easier to maintain a list of known binary file extensions that we want to filter out.

workspace-files/main.go
tybalex: would you prefer a `white list` rather than this black list?

workspace-files/tool.gpt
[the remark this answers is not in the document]
tybalex: agreed. Supports only plain text formats (e.g., .txt, .md) should be enough removed the rest text.

workspace-files/tool.gpt
```diff
@@ -38,14 +38,14 @@ Params: dir: The directory to list files from
...
 Params: filename: The filename to read
 
 #!${GPTSCRIPT_TOOL_DIR}/bin/gptscript-go-tool read
 
 ---
 Name: workspace_write
-Description: Write contents to a file, overwriting the existing file if it exists.
+Description: Writes content to a file, replacing any existing content if the file already exists. Supports only plain text formats (e.g., .txt, .md) and does not allow writing to binary files (e.g., .pdf, .docx, .pptx, .xlsx, .jpg, .png, .gif, .mp3, .mp4, .zip, .rar).
```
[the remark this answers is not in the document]
tybalex: yep. removed

tybalex:
it is very interesting that, now that I prompt the Obot not to read non-plain texts with workspace_read, it actually uses `knowledge ` to find the answer:
Screenshot 2025-03-11 at 3 20 21 PM

workspace-files/main.go
cjellick: no. you convinced me

workspace-files/main.go
cjellick: ok. you convinced me.

cjellick, line 177:
```suggestion
	// Check for common binary file types that are not supported. Less common types should be caught by the utf8.Valid() check
	for _, ext := range unsupportedFileTypes {
```

cjellick (approved):

PR #488, opened 2025-03-12
feat: Google calendar tool

tybalex:
https://github.com/obot-platform/obot/issues/2042

Tools:
- Event:
     - List/Get/Move/Create/Update/Delete Event
     - List Recurring Event instances
- Calendar:
     - List/Create/Get/Update/Delete Calendar


tybalex:
PR is ready for review, but we won't merge it until Acorn's google Oauth App has the scope of google calendar added.

google/calendar/tools/helper.py
```diff
@@ -0,0 +1,66 @@
...
+    except HttpError as err:
+        print(err)
+        exit(1)
+
+
+def get_user_timezone(service):
+    """Fetches the authenticated user's time zone from calendar settings."""
+    try:
+        settings = service.settings().get(setting="timezone").execute()
+        return settings.get("value", "UTC")  # Default to UTC if not found
```
drpebcak, line 58: Does it make sense to default to UTC, or to use `OBOT_USER_TIMEZONE` to get whatever timezone their browser is reporting instead?

tybalex: yeah I'd prefer `OBOT_USER_TIMEZONE` if it's available. is it something I can read from env variable?

drpebcak: yes its available as an env var. here is an example - https://github.com/obot-platform/tools/blob/main/google/search-console/src/keywords.js#L4

drpebcak (approved):

google/calendar/credential/tool.gpt
```diff
@@ -0,0 +1,21 @@
...
+Tools: ../../../oauth2
+
+#!sys.call ../../../oauth2
+
+{
+"oauthInfo": {
+	"integration": "google-calendar",
+	"token": "GOOGLE_OAUTH_TOKEN",
+	"scope": [
+		"https://www.googleapis.com/auth/userinfo.email",
```
StrongMonkey, line 16: @tybalex tbh I haven't tested the scopes. I think you will need to double check on this

StrongMonkey (approved):

google/calendar/tools/helper.py
tybalex: updated to OBOT_USER_TIMEZONE

google/calendar/credential/tool.gpt
tybalex: ok

tybalex: yeah it is not necessary, removed

PR #489, opened 2025-03-12
fix: screenshot download path

njhale:
The path to access workspace files from the UI has changed slightly,
which breaks the links currently produced by the `Screenshot` tool.

Update the download URLs produced by the `Screenshot` tool to account
for the change.

Addresses https://github.com/obot-platform/obot/issues/2059

### DIFF
--- a/browser/src/screenshot.ts
+++ b/browser/src/screenshot.ts
@@ -37,7 +37,7 @@ export async function screenshot (
     const screenshotPath = workspaceId !== undefined ? `files/${screenshotName}` : screenshotName
 
     // Save the screenshot to the workspace
-    await client.writeFileInWorkspace(screenshotPath, screenshot, workspaceId)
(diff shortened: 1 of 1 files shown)


browser/src/screenshot.ts
```diff
@@ -37,7 +37,7 @@ export async function screenshot (
     const screenshotPath = workspaceId !== undefined ? `files/${screenshotName}` : screenshotName
 
     // Save the screenshot to the workspace
-    await client.writeFileInWorkspace(screenshotPath, screenshot, workspaceId)
+    await client.writeFileInWorkspace(screenshotPath, new Uint8Array(screenshot).buffer, workspaceId)
```
njhale, line 40: This isn't a functional change. It stops the linter from complaining.

StrongMonkey (approved):

cjellick (approved):

PR #492, opened 2025-03-12
enhance: turn credential helpers into daemons

g-linville:
requested by @ibuildthecloud 

I tested this with both encryption enabled and disabled. All four credential operations (list, get, store, erase) work.


credential-stores/pkg/common/db.go
g-linville, line 1: this file had the wrong name for a long time so I decided to finally change it

ibuildthecloud (approved):

PR #499, opened 2025-03-12
Fix: fix lastRow bug, add ability to add multiple rows

StrongMonkey:
We have a bug in our existing implementation that when adding multiple rows, the last used row is calcaulated incorrectly so multiple write operation will be overriding existing rows instead of creating new rows.  This PR fixed the issue to properly calculate the last used row according to its address.

Also we should allow updating multiple rows at once to improve the effiefiency.


excel/tool.gpt
```diff
@@ -70,13 +70,13 @@ Param: query: The sql-like query to run against the spreadsheet. Should be the f
...
 Name: Add Worksheet Row
-Description: Adds a row to an existing worksheet in a workbook.
+Description: Adds rows to an existing worksheet in a workbook.
 Share Context: Excel Context
 Credential: ./credential
 Share Tools: List Workbooks, List Worksheets, Get Worksheet Data
 Param: workbook_id: ID of the workbook to add row to
 Param: worksheet_id: ID of the worksheet to add row to
-Param: contents: pipe-separated (`|`) values to add as a new row
+Param: contents: each row is a pipe-separated (`|`) values. Rows are separated by newlines (`\n`).
```
cjellick, line 79:
```suggestion
Param: contents: The rows to add to the worksheet. Cells within a row must be pipe-separated (`|`). Rows must be separated by newlines (`\n`).
```
_Not tested, so see how it does._

g-linville (approved):

cjellick (approved):

PR #518, opened 2025-03-18
Enhance: make sure slack to correctly format message

StrongMonkey:
Make sure slack can format message correctly before sending it.

### DIFF
--- a/slack/tool.gpt
+++ b/slack/tool.gpt
@@ -2,7 +2,7 @@
 Name: Slack
 Description: Tools for interacting with Slack
 Metadata: bundle: true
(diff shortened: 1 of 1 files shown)


slack/tool.gpt
```diff
@@ -192,6 +192,21 @@ Credential: ./credential
 
 #!/usr/bin/env node ${GPTSCRIPT_TOOL_DIR}/index.js userContext
 
+---
+Name: Format Message
```
njhale, line 196: Could this be an input filter instead?

njhale (dismissed):

slack/tool.gpt
StrongMonkey: i have tried input filter and it doesn't work well. https://acorn-io.slack.com/archives/C03FU91U4SX/p1742329698980689

tybalex:
Hey I am sure this implementation will work. But would it be better to use a `markdown-to-slack` package such as https://github.com/tryfabric/mack/tree/master? It is more deterministic and saves a llm call.

StrongMonkey:
@tybalex Good point, I will give it a shot.

tybalex (approved):

PR #526, opened 2025-03-24
fix: use preferred timezone to display outlook message dates

njhale:
Addresses https://github.com/obot-platform/obot/issues/583

Note: The search was working properly in the case explained in the referenced issue, but the dates were being displayed in UTC. This change consistently fixed that issue for me. I also tried an output filter, but it wasn't clear if that actually helped _more_ than just rewording the context.


g-linville (approved):

tybalex (approved):
lgtm. Should we apply this change to other tools such as gmail too?

PR #530, opened 2025-03-26
chore: knowledge: cleanup dependencies

iwilltry42:
This integrates only the necessary chromem-go code and removes all legacy or unused stuff.
Yes, we only use the OpenAI embeddings provider.

Issue: https://github.com/obot-platform/obot/issues/2189


knowledge/pkg/vectorstore/types/types.go
```diff
@@ -1,9 +1,18 @@
 package types
 
+// Disclaimer: a lot of this was taken over from our custom for of github.com/philippgille/chromem-go which deviated quite a lot from the original,
```
g-linville, line 3: nit:

```suggestion
// Disclaimer: a lot of this was taken over from our custom fork of github.com/philippgille/chromem-go which deviated quite a lot from the original,
```

g-linville (approved):

PR #545, opened 2025-03-31
feat: rewrite anthropic-model-provider as openai-style proxy in Go and add thinking mode

iwilltry42:
We cannot replace the Python implementation yet since the anthropic-bedrock provider is still relying on it.

Issue: https://github.com/obot-platform/obot/issues/1999

### DIFF
--- a/anthropic-model-provider-go/main.go
+++ b/anthropic-model-provider-go/main.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httputil"
+	"os"
+
+	aproxy "github.com/obot-platform/tools/anthropic-model-provider/proxy"
+	"github.com/obot-platform/tools/openai-model-provider/proxy"
+)
+
+func main() {
+	isValidate := len(os.Args) > 1 && os.Args[1] == "validate"
+
+	cfg := &proxy.Config{
+		APIKey:          os.Getenv("OBOT_ANTHROPIC_MODEL_PROVIDER_API_KEY"), // optional, as e.g. Ollama doesn't require an API key
+		ListenPort:      os.Getenv("PORT"),
+		BaseURL:         "https://api.anthropic.com/v1/",
+		RewriteModelsFn: aproxy.RewriteModelsResponse,
+		RewriteHeaderFn: func(header http.Header) {
+			header.Del("Authorization")
+			header.Set("x-api-key", os.Getenv("OBOT_ANTHROPIC_MODEL_PROVIDER_API_KEY"))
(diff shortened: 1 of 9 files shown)


anthropic-model-provider-go/main.go
```diff
@@ -0,0 +1,54 @@
...
+		Director: prox.AnthropicProxyRedirect,
+	}
+	reverseProxyModels := &httputil.ReverseProxy{
+		Director:       prox.AnthropicProxyRedirect,
+		ModifyResponse: aproxy.RewriteModelsResponse,
+	}
+	cfg.CustomPathHandleFuncs = map[string]http.HandlerFunc{
+		"/v1/models": reverseProxyModels.ServeHTTP,
+		"/v1/":       reverseProxy.ServeHTTP,
+	}
```
thedadams, line 40: There seem to be a few ways to set the model response re-writer now. After this goes in, and we have some time, someone else should look at that.

anthropic-model-provider-go/proxy/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+type ThinkingRequestBody struct {
+	openai.ChatCompletionRequest
+	Thinking ThinkingConfig `json:"thinking,omitempty"`
+}
+
+func isThinkingModel(model string) bool {
+	if strings.HasSuffix(model, "-thinking") {
+		return true
+	}
+	return false
```
thedadams, line 70: nit:
```suggestion
	return strings.HasSuffix(model, "-thinking")
```

thedadams (approved):

anthropic-model-provider-go/proxy/proxy.go
```diff
@@ -0,0 +1,102 @@
...
+	bodyBytes, err := io.ReadAll(req.Body)
+	if err != nil {
+		fmt.Println("failed to read request body, error: ", err.Error())
+		return
+	}
+
+	var reqBody openai.ChatCompletionRequest
+	if err := json.Unmarshal(bodyBytes, &reqBody); err == nil && needsModification(reqBody) {
+		if err := modifyRequestBody(req, &reqBody); err != nil {
+			fmt.Println("failed to modify request body for o1, error: ", err.Error())
```
g-linville, line 48: this isn't for o1
```suggestion
			fmt.Println("failed to modify request body for claude, error: ", err.Error())
```

g-linville (approved):

anthropic-model-provider-go/proxy/proxy.go
iwilltry42: Copy Pasta 🤷‍♂️

PR #546, opened 2025-04-01
chore: auth providers: use Postgres for session storage

g-linville:
depends on https://github.com/obot-platform/oauth2-proxy/pull/4

I'll fix the go.mod files in this one once that one is merged.


tybalex (approved):

google-auth-provider/tool.gpt
```diff
@@ -12,6 +12,7 @@ Credential: ../placeholder-credential as google-auth-provider
     "icon": "/admin/assets/google_icon_small.png",
     "iconDark": "/admin/assets/google_icon_small.png",
     "link": "https://google.com/",
+    "postgresTablePrefix": "google_",
```
njhale, line 15: We should probably add the underscore in https://github.com/obot-platform/obot/pull/2346/files#diff-1d620309aefdeee90bc0b70186dec69b901dff7d8dae034d0124565d98dc17f6R251 so that the metadata field is the same as the value the auth provider uses for the oauth2-proxy config.

g-linville: fixed here, enterprise tools, and in oauth2-proxy

thedadams (approved):

njhale (approved):

PR #551, opened 2025-04-02
fix: detect userType and return Proper error message when user has no permission

tybalex:
To address issue comment : https://github.com/obot-platform/obot/issues/1940#issuecomment-2754881365

Guest users within a Microsoft organization are restricted from accessing Groups features.

### DIFF
--- a/outlook/mail/pkg/commands/listGroups.go
+++ b/outlook/mail/pkg/commands/listGroups.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/gptscript-ai/tools/outlook/mail/pkg/client"
 	"github.com/gptscript-ai/tools/outlook/mail/pkg/global"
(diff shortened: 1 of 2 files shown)


outlook/mail/pkg/commands/listGroups.go
```diff
@@ -18,6 +19,11 @@ func ListGroups(ctx context.Context) error {
 
 	groups, err := graph.ListGroups(ctx, c)
 	if err != nil {
+		userType, getUserTypeErr := graph.GetUserType(ctx, c)
+		if getUserTypeErr == nil && strings.EqualFold(userType, "Guest") {
+			fmt.Printf("UserType '%s' does not have enough permissions to list groups.\n", userType)
```
thedadams, line 24:
```suggestion
			fmt.Printf("User has type '%s', which does not have enough permissions to list groups.\n", userType)
```

thedadams (approved):

StrongMonkey (approved):

outlook/mail/pkg/commands/listGroups.go
tybalex: 👍

PR #553, opened 2025-04-03
chore: lock dependency versions

drpebcak:


thedadams (approved):
Just to say it out loud: we need to do the same thing in enterprise-tools

g-linville (approved):

PR #556, opened 2025-04-04
fix: Google calendar recurrence rule string

tybalex:
- Update the prompt for Obot --  `json array of strings` instead of `json array`, reduce the possibility of LLM making mistakes. The LLM will now more likely to generate an array like `[ "RRULE:FREQ=YEARLY"]` instead of a string like `"[\"RRULE:FREQ=YEARLY\"]"`
- add one line of code to try to fix bad input.

### DIFF
--- a/google/calendar/tool.gpt
+++ b/google/calendar/tool.gpt
@@ -66,7 +66,7 @@ Param: start_datetime: (Optional) Start date and time of the event to create. Mu
 Param: end_date: (Optional) The date, in the format "yyyy-mm-dd", if this is an all-day event.
 Param: end_datetime: (Optional) End date and time of the event to create. Must be a valid RFC 3339 formatted date/time string. A time zone offset is required unless a time zone is explicitly specified in timeZone.
 Param: attendees: (Optional) A comma separated list of email addresses of the attendees.
-Param: recurrence: (Optional) A JSON array of RRULE, EXRULE, RDATE and EXDATE lines for a recurring event, as specified in RFC5545. Note that DTSTART and DTEND lines are not allowed in this field, because they are already specified in the start_datetime and end_datetime fields.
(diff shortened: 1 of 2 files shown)


google/calendar/tools/event.py
```diff
@@ -417,6 +413,47 @@ def quick_add_event(service):
...
+        ValueError: If the recurrence string is not a valid recurrence rule syntax.
+
+    Returns:
+        list: A list of recurrence rules
+    """
+    try:
+        recurrence_list = json.loads(recurrence)
+    except json.JSONDecodeError:
+        try:
+            fixed_recurrence = recurrence.encode().decode(
```
tybalex, line 434: try to fix the input by treating all the \ escape sequences like they were actual escaped characters again.

thedadams (approved):

google/calendar/tool.gpt
```diff
@@ -88,7 +88,7 @@ Param: end_date: (Optional) The date, in the format "yyyy-mm-dd", if this is an
...
 Param: add_attendees: (Optional) A comma separated list of email addresses of the attendees. This will add the new attendees to the existing attendees list.
 Param: replace_attendees: (Optional) A comma separated list of email addresses of the attendees. This is only valid when add_attendees is empty. This will replace the existing attendees list with the new list.
-Param: recurrence: (Optional) A JSON array of RRULE, EXRULE, RDATE and EXDATE lines for a recurring event, as specified in RFC5545. Note that DTSTART and DTEND lines are not allowed in this field, because they are already specified in the start_datetime and end_datetime fields.
+Param: recurrence: (Optional) For a recurring event, provide a JSON array of strings, where each string is an RRULE, EXRULE, RDATE, or EXDATE line as defined by the RFC5545. For example,["RRULE:FREQ=YEARLY", "EXDATE:20250403T100000Z"]. Note that DTSTART and DTEND lines are not allowed in this field, because they are already specified in the start_datetime and end_datetime fields.
```
StrongMonkey, line 91: Just another thought, i feel like sometime ask LLM to return simplified result like comma separated strings then in code construct those to json array might be less error-prune.

StrongMonkey (approved):

google/calendar/tool.gpt
tybalex: Yeah good call — I thought about that too, but RRULEs can have commas inside (like BYDAY=MO,TU,WE), so splitting gets messy.

PR #565, opened 2025-04-09
chore: move tool build into tools repo

drpebcak:


.dockerignore
```diff
@@ -0,0 +1,7 @@
+**/node_modules/
+.local/
+.git/
+**/.venv
+.github/
+.cursor/
```
thedadams, line 6: nit: might as well add .vscode and .idea here.

njhale (approved):

PR #568, opened 2025-04-09
Fix: fix the ability in discord to reply in thread

StrongMonkey:
End up adding support to reply in thread in discord.

### DIFF
--- a/discord/src/tools/sendMessageInThread.ts
+++ b/discord/src/tools/sendMessageInThread.ts
@@ -1,52 +1,63 @@
-import { TextChannel, ThreadChannel } from 'discord.js';
-import { client } from '../client.js';
-import { formatTime } from '../utils.js';
+import {
+  Message,
+  PublicThreadChannel,
+  TextChannel,
+  ThreadChannel,
+} from "discord.js";
+import { client } from "../client.js";
+import { formatTime } from "../utils.js";
 
 export async function sendMessageInThread() {
-  if (!process.env.CHANNELID || !process.env.GUILDID || !process.env.THREADID || !process.env.TEXT) {
-    throw new Error('CHANNELID, GUILDID, THREADID, and TEXT environment variables are required for sendMessageInThread');
+  if (
+    !process.env.CHANNELID ||
+    !process.env.GUILDID ||
+    !process.env.MESSAGEID ||
+    !process.env.TEXT
+  ) {
+    throw new Error(
+      "CHANNELID, GUILDID, MESSAGEID, and TEXT environment variables are required for sendMessageInThread"
+    );
   }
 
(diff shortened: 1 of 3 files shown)


discord/tool.gpt
```diff
@@ -106,8 +106,8 @@ Share Tools: List Channels, Search Channels, Get Channel History, Get Thread His
 Share Context: Discord Context
 Param: channelid: the ID of the channel containing the thread
 Param: guildid: the ID of the guild that the channel belongs to
-Param: threadid: the ID of the thread to send the message to
 Param: text: the text content to send
+Param: messageid: the ID of the message to create a thread on
```
njhale, line 110: Seems like this should be marked as "optional", since it doesn't look like it's necessary to send a new message on a thread.

discord/src/tools/sendMessageInThread.ts
```diff
@@ -1,52 +1,63 @@
...
+import { formatTime } from "../utils.js";
 
 export async function sendMessageInThread() {
-  if (!process.env.CHANNELID || !process.env.GUILDID || !process.env.THREADID || !process.env.TEXT) {
-    throw new Error('CHANNELID, GUILDID, THREADID, and TEXT environment variables are required for sendMessageInThread');
+  if (
+    !process.env.CHANNELID ||
+    !process.env.GUILDID ||
+    !process.env.MESSAGEID ||
+    !process.env.TEXT
```
njhale, line 15: Is `MESSAGEID` really required if `CHANNELID` resolves to a channel of type `ThreadChannel`?

njhale (changes requested):

g-linville (approved):
I'm good with this as long as it works in both of these ways:

- Obot is pinged inside of an existing thread, and responds in the thread
- Obot is pinged in a new message in a normal channel, and responds by creating a new thread from that message and posting a message in that new thread

It's not clear to me from the code if the first case is still supported here, but it needs to be.

StrongMonkey:
@g-linville yeah, the first case works because channel id is basically thread id when user sends message inside a thread

discord/tool.gpt
StrongMonkey: yeah, I am afraid that optional might break it for new message. We should always have message id in context when replying in the thread. although it is not used inside thread.

discord/src/tools/sendMessageInThread.ts
StrongMonkey: Not needed, but the context should always have the MESSAGEID if replying in a thread.We don't really need message id if it is replying inside the thread rather than creating new thread, but having LLM providing MESSAGEID will make the behavior more predicable.

njhale (approved):

PR #576, opened 2025-04-11
Enhance: Add abilities to set typing status in slack

StrongMonkey:
Add ability to set typing status in our slack app to make it fun.

### DIFF
--- a/slack/tool-bot.gpt
+++ b/slack/tool-bot.gpt
@@ -2,7 +2,7 @@
 Name: Slack Bot
 Description: Tools for interacting with Slack as an App
 Metadata: bundle: true
-Share Tools: Send Direct Message As Bot, List Channels As Bot, Search Channels As Bot, Get Channel History As Bot, Get Channel History By Time As Bot, Get Thread History From Link As Bot, Get Thread History As Bot, Send Message As Bot, Send Message in Thread As Bot, List Users As Bot, Search Users As Bot, Send DM in Thread As Bot, Send Direct Message As Bot, Get Message Link As Bot, Get DM History As Bot, Get DM Thread History As Bot,  Format Message from ./tool.gpt
+Share Tools: Send Direct Message As Bot, List Channels As Bot, Search Channels As Bot, Get Channel History As Bot, Get Channel History By Time As Bot, Get Thread History From Link As Bot, Get Thread History As Bot, Send Message As Bot, Send Message in Thread As Bot, List Users As Bot, Search Users As Bot, Send DM in Thread As Bot, Send Direct Message As Bot, Get Message Link As Bot, Get DM History As Bot, Get DM Thread History As Bot, Send Typing Event As Bot, Format Message from ./tool.gpt
 
 ---
 Name: List Channels As Bot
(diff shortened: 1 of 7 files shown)


slack/tool-bot.gpt
```diff
@@ -173,6 +173,17 @@ Param: limit: the number of messages to return
...
 #!/usr/bin/env node ${GPTSCRIPT_TOOL_DIR}/index.js getDMThreadHistory
 
+---
+Name: Send Typing Event As Bot
+Description: Send a typing event to a channel in the Slack workspace as a bot
+Credential: ./credential/bot.gpt
+Share Tools: List Channels As Bot, Search Channels As Bot
+Param: channelid: the ID of the channel to send the typing event to
+Param: threadid: the ID of the thread to send the typing event to
+Param: status: the status to set the typing event to
```
g-linville, line 183: I assume there are only a few valid values for this param? They should be listed here and in the other gpt file.

g-linville (changes requested):

tybalex (approved):
is threadID optional though?

slack/tool-bot.gpt
StrongMonkey: status? no, you can set anything

StrongMonkey:
>is threadID optional though?

Required in api

slack/tool-bot.gpt
tybalex: lol wait so the status can be `Puppy is making up things...`?

StrongMonkey: it can even have emoji

tybalex: that would be funny.. we should set that as default 🤣

tybalex: no just kidding

tybalex:
actually, correct me if I'm wrong, should this be a tool? I guess why don't we simply invoke this method to set the status, before calling chat completion?

StrongMonkey:
@tybalex Hmm, I think it is better to have it be a tool so it is more customziable from user to decide when to set typing status. We don't want to this to be integrated into obot

slack/tool-bot.gpt
g-linville: @StrongMonkey oh that's cool that we can make it be whatever. I think the param description should explain that this is the typing status that will be shown to the slack users. I find the parameter confusing as it is.

tybalex:
> @tybalex Hmm, I think it is better to have it be a tool so it is more customziable from user to decide when to set typing status. We don't want to this to be integrated into obot

oh yeah that makes sense.

slack/tool-bot.gpt
StrongMonkey: yeah, good point, updated.

g-linville (approved):

slack/tool-bot.gpt
StrongMonkey: I have tweaked the prompt a bit to allow emoji..looks like if you say it is a status in slack thread sometime LLM will not allow you to add emoji to it.

PR #583, opened 2025-04-15
enhance: improve error messages for credential validation tools

g-linville:
Thanks to @StrongMonkey for the design here. This introduces a change to allow us to bubble a structured error all the way back up to obot for credentials with a `validationTool` set.

for https://github.com/obot-platform/obot/issues/2367

### DIFF
--- a/generic-credential/main.go
+++ b/generic-credential/main.go
@@ -71,7 +71,7 @@ func mainErr() error {
 	}
 
 	if err = validateCredential(ctx, client, in.ValidationTool, envs); err != nil {
-		return fmt.Errorf("error validating credentials: %w", err)
+		return err
 	}
 
 	b, err := json.Marshal(map[string]any{"env": envs})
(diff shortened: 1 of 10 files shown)


StrongMonkey (approved):

generic-credential/main.go
```diff
@@ -163,11 +163,18 @@ func validateCredential(ctx context.Context, client *gptscript.GPTScript, tool s
...
 	if err != nil {
 		errStr, _, _ := strings.Cut(err.Error(), ": exit status ")
 		return errors.New(errStr)
 	}
 
+	var errResp struct {
+		Error string `json:"error"`
+	}
+	if err := json.Unmarshal([]byte(output), &errResp); err == nil && errResp.Error != "" {
+		return errors.New(errResp.Error)
```
thedadams, line 176: Just to make sure: based on the changes in the other languages, I expect that `errResp.Error` is already JSON?

g-linville: It's normally just going to be a string. The `validationTool` tools, if they error, will print something in the format `{"error": "this is the error message"}` and then `exit(0)`, which is what we are looking for here.

thedadams (approved):

PR #586, opened 2025-04-16
feat: add new virus total provider for file scanning

thedadams:


njhale (approved):

thedadams:
I added another, faster file scanner. Re-requesting reviews.

meta-defender-provider/main.go
```diff
@@ -0,0 +1,54 @@
...
+
+func main() {
+	apiKey := os.Getenv("META_DEFENDER_API_KEY")
+	if apiKey == "" {
+		fmt.Println("META_DEFENDER_API_KEY is not set")
+		os.Exit(1)
+	}
+
+	if len(os.Args) > 1 && os.Args[1] == "validate" {
+		if err := validate(context.Background(), apiKey); err != nil {
```
StrongMonkey, line 20: grant changed the framework so that if you are rteurning structure error `{"error": err.string()}` it should print the extra error back to UI without extra error wrapping.

virus-total-provider/filehandler.go
```diff
@@ -0,0 +1,225 @@
...
+	Category      string `json:"category"`
+	EngineName    string `json:"engine_name"`
+	EngineVersion string `json:"engine_version"`
+	EngineUpdate  string `json:"engine_update"`
+	Method        string `json:"method"`
+	Result        string `json:"result"`
+}
+
+type Stats struct {
+	ConfirmedTimeout int `json:"confirmed-timeout"`
```
StrongMonkey, line 217: do they really have both `-` and `_` in their api response....

virus-total-provider/filehandler.go
```diff
@@ -0,0 +1,225 @@
...
+		"file": bytes.NewReader(contents),
+	}, &resp)
+}
+
+func (f *fileHandler) pollForResult(ctx context.Context, analysisID string) (bool, error) {
+	var (
+		resp analysisResult
+		err  error
+	)
+	ticker := time.NewTicker(time.Second)
```
StrongMonkey, line 78: want to make sure, polling every second will be not aggressive in their api?

thedadams: Yeah, I don't know. Should we do something different do you think?

virus-total-provider/filehandler.go
thedadams: Yep. This is from their docs:
```
{
  "data": {
    "attributes": {
    	 "date": <int:timestamp>,
       	"results": {
          	"<string>": {
            	"category": "<string>",
            	"engine_name": "<string>",
            	"engine_version": "<string>",
            	"engine_update": "<string>",
            	"method": "<string>",
            	"result": "<string>"
        	},
        	"stats": {
            	"confirmed-timeout": <int>,
           		"failure": <int>,
          		"harmless": <int>,
          		"malicious": <int>,
          		"suspicious": <int>,
          		"timeout": <int>,
           		"type-unsupported": <int>,
          		"undetected": <int>
        	},
        	"status": "<string>"      
    },
    "id": "<string>",
    "type": "analysis"
  }
}
```

virus-total-provider/filehandler.go
StrongMonkey: Oh, I was thinking maybe like every 5/10 second to make sure we don't get rate limited. But it is not a problem then that's probably fine.

meta-defender-provider/main.go
thedadams: Updated.

StrongMonkey: One last thing, it has to exit with 0 so that gptscript will not report extra error wrapping. The framework will catch this special error and throw an error properly.

StrongMonkey (approved):

meta-defender-provider/main.go
thedadams: Done. Thanks, Daishan!

PR #587, opened 2025-04-16
Enhance: Refactored getAttachment tool to download the file instead

StrongMonkey:
Instead of reading the file from attachment and put it directly into context, we want to download the attachment file and put it directly into workspace and as a knowledge file. This follows the same pattern when user uploads a file in UI where we save the file into workspace and also add it as knowledge file so that we have the context in the chat.

https://github.com/obot-platform/obot/issues/2563

require: https://github.com/obot-platform/obot/pull/2562


njhale (approved):

StrongMonkey:
I pushed a new commit to add Read Attachment tool per discussion.

PR #588, opened 2025-04-16
Fix: Add abilities to list other labels and select category

StrongMonkey:
Add abilities to list primary tab from gmail by default. Also add abilities to list other labels, like sent, trash, spam.

### DIFF
--- a/google/gmail/listInbox.py
+++ b/google/gmail/listInbox.py
@@ -8,7 +8,12 @@ async def main():
     max_results = os.getenv('MAX_RESULTS', '100')
     if max_results is not None:
         max_results = int(max_results)
-    query = 'label:inbox'
+    category = os.getenv('CATEGORY', 'primary')
+    label = os.getenv('LABEL', 'inbox')
+    if label != 'inbox':
+        query = f'label:{label}'
+    else:
+        query = f'label:{label} category:{category}'
 
     service = client('gmail', 'v1')
     await list_messages(service, query, max_results)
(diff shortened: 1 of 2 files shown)


mighty-pup-acorn:
### Enhancement Proposal

This pull request is related to an enhancement proposal to add abilities to list other labels such as sent, trash, and spam, in addition to the primary tab from Gmail by default. This enhancement aims to improve user experience by providing more comprehensive email management capabilities.

#### Proposed Solution
- Implement functionality to list additional labels like sent, trash, and spam.
- Ensure seamless integration with the existing system to maintain user experience.

#### Benefits
- Improved email management by allowing users to access more labels.
- Enhanced user experience with more options for email categorization.

#### Acceptance Criteria
- [ ] Ability to list additional labels is implemented and functional.
- [ ] Integration with existing systems is seamless and does not disrupt current functionalities.
- [ ] User feedback is positive regarding the new label management capabilities.

google/gmail/listInbox.py
```diff
@@ -8,7 +8,12 @@ async def main():
     max_results = os.getenv('MAX_RESULTS', '100')
     if max_results is not None:
         max_results = int(max_results)
-    query = 'label:inbox'
+    category = os.getenv('CATEGORY', 'primary')
+    label = os.getenv('LABEL', 'inbox')
```
tybalex, line 12: Can we validate that the values of category and label are among the allowed options?
for example : `if "label" is not in ['inbox', 'starred', 'important', 'sent', 'drafts', 'trash', 'spam', 'all']`

njhale: We have a set of options here, but gmail allows you to apply custom labels too (I use the feature to pretty extensively myself), so it may make sense to make this less restrictive as well.

njhale (approved):

google/gmail/listInbox.py
tybalex: I see.  if custom labels are allowed then we don't need to validate it here.

tybalex (approved):

PR #591, opened 2025-04-17
feat: add memory tools

njhale:
Add the `Memory` bundle to provide tools for managing project scoped memory.

Addresses part of https://github.com/obot-platform/obot/issues/2276

Depends on https://github.com/obot-platform/obot/pull/2586


thedadams (approved):

tybalex (approved):
code looks good.
why do we need `list memory` exposed as a tool though?  isn't memory always appended in system prompt.

memory/main.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil
+	}
+
+	if url == "" {
+		url = "http://localhost:8080/api"
+	} else {
+		url += "/api"
+	}
+
+	client := &apiclient.Client{
```
StrongMonkey, line 106: you can actually use apiclient.NewClientFromEnv

StrongMonkey (approved):

njhale:
> why do we need list memory exposed as a tool though? isn't memory always appended in system prompt.

@tybalex the memory is added to the context (system prompt) because the list_memories tool is a `Shared Context` on the `memory_context` tool. We can refactor it so that it's all done in the go code, and there's a `context` command instead of a `list` command, but as it stands, I don't believe the list_memories tool will be exposed for use to the LLM or be visible in the user UI.

memory/main.go
njhale: didn't see this until after I clicked merge. Good call, I'll open a quick follow up to fix in a bit

StrongMonkey: nbd

PR #598, opened 2025-04-21
fix: only include stream options if stream is true

thedadams:


thedadams:
Issue: https://github.com/obot-platform/obot/issues/2660

PR #599, opened 2025-04-22
enhance: add loop-data tool

g-linville:


loop-data/tool.gpt
```diff
@@ -0,0 +1,8 @@
+Name: Loop Data
+Description: Please provide either the data_list or dataset_id parameter, but not both.
```
cjellick, line 2:
```suggestion
Description: Provide either the data_list or dataset_id parameter, but not both.
```

cjellick: What do you think - we're made of tokens? 😝

cjellick (approved):

index.yaml
```diff
@@ -99,6 +99,8 @@ tools:
     reference: ./proxycurl
   memory:
     reference: ./memory
+  loop-data:
+    reference: ./loop-data
```
thedadams, line 103: Does it make sense for this to be in the `system` section?

loop-data/tool.gpt
```diff
@@ -0,0 +1,8 @@
+Name: Loop Data
+Description: Provide either the data_list or dataset_id parameter, but not both.
+Metadata: index: false
```
thedadams, line 3: I think if we move this tool to the `system` section of `index.yaml`, then we can get rid of this.

index.yaml
g-linville: Yeah, it sadly cannot be in the `system` section, because I get an error from the controllers when we try to run the step with it added as one of the tools. I guess system tools can't be handed directly to the LLM, at least through normal runs.

g-linville: Fixed

njhale (approved):

thedadams (approved):

PR #603, opened 2025-04-22
chore: merge m365 creds into one

drpebcak:
relates to https://github.com/obot-platform/obot/issues/2638


microsoft365/credential/tool.gpt
```diff
@@ -1,18 +1,25 @@
-Name: Outlook Calendar OAuth Credential
-Share Credential: outlook-cal-cred as outlook.calendar.write
+Name: Microsoft365 OAuth Credential
+Share Credential: m365-cred as microsoft365.write
```
thedadams, line 2: My only concern here is that we are effectively deleting everyone's credential with this. Is that worth a standup discussion?

thedadams (approved):

tybalex (approved):

PR #606, opened 2025-04-22
enhance: outlook mail: add tools for attachments

g-linville:
for https://github.com/obot-platform/obot/issues/2647

### DIFF
--- a/outlook/mail/pkg/graph/attachments.go
+++ b/outlook/mail/pkg/graph/attachments.go
@@ -1,0 +1,79 @@
+package graph
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	abstractions "github.com/microsoft/kiota-abstractions-go"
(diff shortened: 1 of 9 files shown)


outlook/mail/pkg/graph/attachments.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Check if text is too large by counting tokens
+	tokenizer, err := tiktoken.EncodingForModel("gpt-4")
+	if err != nil {
+		return "", fmt.Errorf("failed to create tokenizer: %w", err)
+	}
+
+	tokens := tokenizer.Encode(result, nil, nil)
+	if len(tokens) > 10000 {
+		return "", fmt.Errorf("attachment content is too large (over 10k tokens)")
```
tybalex, line 75: if it's too large, should we show first 5k/10k tokens and maybe share a link instead?

g-linville: I'm preserving existing behavior here. This was refactored and copied from elsewhere. We could consider changing this.

cjellick (approved):

PR #607, opened 2025-04-23
quick fix: list calendar for users without groups access

tybalex:
https://github.com/obot-platform/obot/issues/2694
personal accounts have no access to groups feature.


outlook/calendar/pkg/graph/calendar.go
```diff
@@ -82,22 +82,20 @@ func ListCalendars(ctx context.Context, client *msgraphsdkgo.GraphServiceClient)
...
-	for _, group := range memberOf.GetValue() {
-		result, err := client.Groups().ByGroupId(util.Deref(group.GetId())).Calendar().Get(ctx, nil)
-		if err != nil {
-			// Some groups don't have calendars and will just error out. That's fine.
-			continue
+	if err == nil { // if user's account has access to groups feature
+		for _, group := range memberOf.GetValue() {
+			result, err := client.Groups().ByGroupId(util.Deref(group.GetId())).Calendar().Get(ctx, nil)
+			if err != nil {
+				// Some groups don't have calendars and will just error out. That's fine.
```
StrongMonkey, line 89: Hmm what is the actual fix? look like the code will do the same behavior?

tybalex: except the logic here:
```
if err != nil {
		return nil, fmt.Errorf("failed to get group memberships: %w", err)
	}
```
Personal accounts do not have access to group features, so this should not be treated as an error.

thedadams (approved):

drpebcak (approved):

PR #608, opened 2025-04-23
Fix: use proper knowledge load flow when loading

StrongMonkey:
When using knowledge to load files for attachements, we don't have a correct flow files we used in standard knowledge so it won't be able to process other file types like pptx. This PR fixed it.

https://github.com/obot-platform/obot/issues/2624


tybalex (approved):

tybalex:
change looks good. Wouldn't it need the `OPENAI_MODEL` env variable for OpenAI OCR model though? I have to set it in another use case where I invoke `knowledge load`

thedadams (approved):

PR #622, opened 2025-04-25
chore: refactor ci

drpebcak:


scripts/package-tools.sh
```diff
@@ -1,177 +1,56 @@
...
-            ./scripts/build.sh
-            echo "Build script for ${REPO} complete!"
-          )
-        else
-            echo "No build script found in ${REPO}"
-        fi
-
-        OBOT_SERVER_VERSIONS="$(cat <<VERSIONS
-"${REPO}": "$(cd "${REPO_NAME}" && git rev-parse --short HEAD)"
+REPO=github.com/obot-platform/tools
```
cjellick, line 4: all the same questions from the enterprise-tools repo regarding cutting out cruft in this file

scripts/package-providers.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
```
cjellick, line 1: not really going to pick this one apart since you said you want to do a separate refactor of it

cjellick (approved):

.github/workflows/build-providers.yaml
```diff
@@ -0,0 +1,57 @@
...
+      cancel-in-progress: true
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ secrets.GHCR_USERNAME }}
+          password: ${{ secrets.GHCR_TOKEN }}
```
thedadams, line 27: Do we need this? Can we use the token passed to the action?

.github/workflows/build-tools.yaml
thedadams, line 1: Same question here about using the GitHub token passed to the action instead of a secret.

scripts/package-providers.sh
```diff
@@ -0,0 +1,81 @@
...
+set -e -x -o pipefail
+
+BIN_DIR=${BIN_DIR:-./bin}
+
+cd /obot-tools
+
+if [ ! -e workspace-provider ]; then
+    git clone --depth=1 https://github.com/gptscript-ai/workspace-provider
```
thedadams, line 9: nit: would be nice to be able to change these with an env var or something.

thedadams (approved):

scripts/package-providers.sh
drpebcak: yeah ive got that marked down as a future enhancement

.github/workflows/build-providers.yaml
drpebcak: hmm.. we probably could get rid of these yeah. ill follow up on that in a separate set of prs

PR #623, opened 2025-04-25
fix: update generate image buffer handling for node 22

njhale:
Addresses https://github.com/obot-platform/obot/issues/2704


tybalex (approved):

njhale:
Closing out b/c the branch name breaks image builds (need to fix that at some point)

PR #625, opened 2025-04-26
fix: Gmail bug archive

tybalex:
- check if label is "" -- to address https://github.com/obot-platform/obot/issues/2727
- allow apply labels update actions to threads for https://github.com/obot-platform/obot/issues/2591#issuecomment-2830543594

### DIFF
--- a/google/gmail/apis/messages.py
+++ b/google/gmail/apis/messages.py
@@ -29,11 +29,13 @@ CATEGORY_LABELS = {
     "CATEGORY_FORUMS": "forums",
 }
 
+
 def modify_message_labels(
     service,
     message_id,
     add_labels=None,
     remove_labels=None,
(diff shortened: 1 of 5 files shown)


google/gmail/apis/messages.py
```diff
@@ -113,12 +115,61 @@ def modify_message_labels(
...
-        .execute()
-    )
+    if apply_action_to_thread:
+        try:
+            thread = get_thread_with_message_id(service, message_id)
+            thread_id = thread.get("id")
+            message_ids = [msg["id"] for msg in thread.get("messages", [])]
+            if message_ids:
+                body["ids"] = message_ids
+                service.users().messages().batchModify(
```
tybalex, line 125: batch operation to all messages in the thread

StrongMonkey (approved):

PR #635, opened 2025-04-29
fix: Outlook mail get groupmail detail

tybalex:
https://github.com/obot-platform/obot/issues/2754
To get email details of group mails

### DIFF
--- a/microsoft365/outlook/mail/pkg/commands/getEmailDetails.go
+++ b/microsoft365/outlook/mail/pkg/commands/getEmailDetails.go
@@ -17,7 +17,7 @@ func GetEmailDetails(ctx context.Context, emailID, groupID, threadID string) err
 	if err != nil {
 		return fmt.Errorf("failed to create client: %w", err)
 	}
-	if groupID != "" {
+	if groupID == "" { // Personal mailbox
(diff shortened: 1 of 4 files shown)


g-linville (approved):

microsoft365/outlook/mail/pkg/commands/getEmailDetails.go
```diff
@@ -12,32 +12,60 @@ import (
...
+			}
+		}
+
+		if sender := result.GetSender(); sender != nil {
+			if email := sender.GetEmailAddress(); email != nil {
+				fmt.Printf("Sender Name: %s\n", util.Deref(email.GetName()))
+				fmt.Printf("Sender Address: %s\n", util.Deref(email.GetAddress()))
+			}
+		}
 	}
```
njhale, line 68: nit: looks like we don't actually need `else`, and this can be de-indented.

njhale (approved):

microsoft365/outlook/mail/pkg/commands/getEmailDetails.go
tybalex: ok

PR #636, opened 2025-04-30
enhance: update model providers to support personal tokens

thedadams:
The model providers have been changed to support user's personal tokens. If configuration for the environment variables has been provided via special headers, then the model providers will use that configuration instead of the "global" configuration. Additionally, the model providers have been changed to be able to start up without configuration so that they can be used only with a user's individual configuration.


cjellick:
Will this be backwards compatible with the prod obot branch or will we need to backport something?

thedadams:
It's backwards compatible.

njhale (approved):

PR #637, opened 2025-04-30
enhance: filter existing memories on create

njhale:
- Add an in-tool call to gptscript to filter out memories that already exist in the set of current memories
- Bump memory obot apiclient go module to get memory API changes

Depends on https://github.com/obot-platform/obot/pull/2764


tybalex (approved):

njhale:
Going to close this out since the chat functionality has been removed in the incarnation this would have been useful in.

PR #643, opened 2025-05-08
Update SmartThings tool.gpt

billabongrob:
Currently says Zoom and that's not correct.


cloudnautique (approved):

cjellick:
we ended up going through and updating all our tool descriptions and hit this in that pr. thanks for pointing it out

PR #649, opened 2025-05-12
Fix: increase the timeout when loading page

StrongMonkey:
The timeout seems to be too low for loading a page. Some page needs to render js to get the full body. Increasing the timeout fixed the issue. https://github.com/obot-platform/obot/issues/2749


njhale (approved):

StrongMonkey:
the rest are formmatter changes.

PR #652, opened 2025-05-13
Fix: add model for image generation

StrongMonkey:
https://github.com/obot-platform/obot/issues/2678

Fix xai model provider that doesn't have image-generation usage.

### DIFF
--- a/openai-model-provider/proxy/rewrite.go
+++ b/openai-model-provider/proxy/rewrite.go
@@ -117,3 +117,61 @@ func RewriteAllModelsWithUsage(usage string, filters ...func(string) bool) func(
 		return nil
 	}
 }
+
+// RewriteAllModelsWithUsageMap returns a response modifier that marks all models with the specified usage
+func RewriteAllModelsWithUsageMap(usageMap map[string][]func(string) bool) func(*http.Response) error {
+	return func(resp *http.Response) error {
+		if resp.StatusCode != http.StatusOK {
+			return nil
+		}
+
+		defer resp.Body.Close()
+
+		var body io.Reader = resp.Body
+		if resp.Header.Get("Content-Encoding") == "gzip" {
+			gzReader, err := gzip.NewReader(resp.Body)
+			if err != nil {
+				return fmt.Errorf("failed to create gzip reader: %w", err)
+			}
+			defer gzReader.Close()
+			resp.Header.Del("Content-Encoding")
+			body = gzReader
+		}
+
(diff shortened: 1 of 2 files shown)


openai-model-provider/proxy/rewrite.go
```diff
@@ -117,3 +117,59 @@ func RewriteAllModelsWithUsage(usage string, filters ...func(string) bool) func(
...
+					for _, filter := range filters {
+						if filter(model.ID) {
+							model.Metadata["usage"] = usage
+							break
+						}
+					}
+				}
+				models.Data[i] = model
+			}
+		}
```
g-linville, line 164: The design of this could technically allow for one model to have the usage set to one thing, and then have it get overwritten by a later iteration if it matches the filters for a different usage. Do we want to do that? Or should we add something like this to the loop to skip models that already have a usage?

```go
if model.Metadata["usage"] != "" {
    continue
}
```

g-linville (approved):

openai-model-provider/proxy/rewrite.go
StrongMonkey: For xai, it doesn't have usage. So we are just grouping usage based on its own name.

g-linville: Yeah I was just thinking for the case of reusability, since this is in the package that the other model providers depend on

StrongMonkey: Added the skip logic you suggested so that if it has usage then it will skip.

thedadams (approved):
It looks like this logic could be achieved with the existing `RewriteAllModelsWithUsage` function. But your implementation seems simpler, so I am good with it.

openai-model-provider/proxy/rewrite.go
```diff
@@ -117,3 +117,63 @@ func RewriteAllModelsWithUsage(usage string, filters ...func(string) bool) func(
...
+
+		for usage, filters := range usageMap {
+			for i, model := range models.Data {
+				if model.Metadata == nil {
+					model.Metadata = make(map[string]string)
+				}
+
+				if model.Metadata["usage"] != "" {
+					continue
+				}
```
g-linville, line 154: tiny optimization

```suggestion
				if model.Metadata == nil {
					model.Metadata = make(map[string]string)
				} else if model.Metadata["usage"] != "" {
					continue
				}
```

g-linville (approved):

PR #658, opened 2025-05-15
fix: anthropic models fix when tool call with no args

tybalex:
https://github.com/obot-platform/obot/issues/2936
`Arguments` cannot be omitted for Anthropic models

### DIFF
--- a/anthropic-model-provider-go/proxy/proxy.go
+++ b/anthropic-model-provider-go/proxy/proxy.go
@@ -91,11 +91,9 @@ func isThinkingModel(model string) bool {
 
 func modifyRequestBody(reqBody *openai.ChatCompletionRequest) (any, bool) {
 	for i := range reqBody.Messages {
-		if reqBody.Messages[i].ToolCalls != nil {
-			for j := range reqBody.Messages[i].ToolCalls {
-				if reqBody.Messages[i].ToolCalls[j].Function.Arguments == "" {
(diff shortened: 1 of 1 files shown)


anthropic-model-provider-go/proxy/proxy.go
```diff
@@ -90,6 +90,15 @@ func isThinkingModel(model string) bool {
...
 
 func modifyRequestBody(reqBody *openai.ChatCompletionRequest) (any, bool) {
+	for i := range reqBody.Messages {
+		if reqBody.Messages[i].ToolCalls != nil {
+			for j := range reqBody.Messages[i].ToolCalls {
+				if reqBody.Messages[i].ToolCalls[j].Function.Arguments == "" {
+					reqBody.Messages[i].ToolCalls[j].Function.Arguments = "{}" // anthropic requires a valid json object
+				}
+			}
+		}
```
thedadams, line 100: Checking for a nil array when looping in Go is not necessary.
```suggestion
		for j := range reqBody.Messages[i].ToolCalls {
			if reqBody.Messages[i].ToolCalls[j].Function.Arguments == "" {
				reqBody.Messages[i].ToolCalls[j].Function.Arguments = "{}" // anthropic requires a valid json object
			}
		}
```

thedadams (approved):

g-linville (approved):

PR #669, opened 2025-05-15
enhance: outlook mail: add tool to download files from onedrive share links

g-linville:
for https://github.com/obot-platform/obot/issues/2625

Since these "OneDrive attachments" that can appear in emails are actually just OneDrive share links, I added a tool to the Outlook Mail bundle to download them, the same way we download normal files in the OneDrive tools.

### DIFF
--- a/microsoft365/outlook/mail/main.go
+++ b/microsoft365/outlook/mail/main.go
@@ -6,9 +6,9 @@ import (
 	"os"
 	"strings"
 
+	"github.com/obot-platform/obot/apiclient"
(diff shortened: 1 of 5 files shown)


tybalex (approved):

tybalex:
should this tool belong to Onedrive Tool bundle though? I guess technically Calendar Tool can return onedrive share links too and we probably don't want to create another `DownloadOnedrivelink` in Calendar Tools?

g-linville:
@tybalex we might want to move it somewhere else in the future, idk. For now I think it just makes the most sense to keep it here, based on the issue I was working on.

PR #673, opened 2025-05-16
fix: Outlook calendar list events

tybalex:
- list events order by time: https://github.com/obot-platform/obot/issues/2741
- list events support `calendar_ids` and `limit`. So it doesn't always have to return events from all calendars. https://github.com/obot-platform/obot/issues/2747

### DIFF
--- a/microsoft365/outlook/calendar/pkg/commands/listEvents.go
+++ b/microsoft365/outlook/calendar/pkg/commands/listEvents.go
@@ -22,7 +22,6 @@ func ListEvents(ctx context.Context, calendarIDstring string, start, end time.Ti
 	if limit == "" {
 		limitInt = 50 // default limit
 	} else {
-		var err error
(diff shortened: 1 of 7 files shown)


microsoft365/outlook/calendar/pkg/commands/listEvents.go
```diff
@@ -15,7 +17,19 @@ import (
 	"github.com/obot-platform/tools/microsoft365/outlook/common/id"
 )
 
-func ListEvents(ctx context.Context, start, end time.Time) error {
+func ListEvents(ctx context.Context, calendarIDstring string, start, end time.Time, limit string) error {
+	var limitInt int32
+	if limit == "" {
+		limitInt = 50 // default limit
+	} else {
+		var err error
```
g-linville, line 25: This line looks unnecessary, since err is redeclared the line below.

```suggestion
```

g-linville (approved):

PR #682, opened 2025-05-19
fix: add proper suffix to downloaded google workspace files

tybalex:
https://github.com/obot-platform/obot/issues/2839

### DIFF
--- a/google/drive/apis/files.py
+++ b/google/drive/apis/files.py
@@ -1,4 +1,4 @@
-from typing import Optional, List
+from typing import Optional, List, Tuple
 from googleapiclient.discovery import Resource
 from googleapiclient.errors import HttpError
 from googleapiclient.http import MediaIoBaseUpload, MediaIoBaseDownload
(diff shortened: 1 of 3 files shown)


cjellick (approved):

google/drive/apis/files.py
```diff
@@ -320,6 +322,19 @@ def download_file(service: Resource, file_id: str) -> Optional[bytes]:
...
+                "application/pdf": ".pdf",
+                "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet": ".xlsx",
+                "application/vnd.openxmlformats-officedocument.wordprocessingml.document": ".docx",
+                "application/vnd.openxmlformats-officedocument.presentationml.presentation": ".pptx",
+            }
+
+            # Add extension if the file doesn't already have it
+            extension = extension_map.get(export_mime_type, ".pdf")
+            if not file_name.lower().endswith(extension):
+                file_name = f"{file_name}{extension}"
```
g-linville, line 336: Is there some kind of MIME type-to-file extension library we could use instead? It seems weird to limit it to only these four types (and to default to PDF).

tybalex: it's a subset of https://developers.google.com/workspace/drive/api/guides/ref-export-formats. I can't find a library for it.

PR #693, opened 2025-06-11
chore: remove bespoke tools and replace with official/community versions

drpebcak:


drpebcak:
I'm assuming that the relevant entries in the obot catalog dir are all we need to replace these tools.

g-linville (approved):

njhale (approved):

PR #694, opened 2025-06-11
chore: replace die-roller tool with an mcp server

drpebcak:


drpebcak:
So now the question is, how do I get this mcp server into obot?

njhale (approved):

PR #699, opened 2025-06-17
feat: google calendar mcp server

tybalex:
replace legacy google calendar tool bundle with the MCP server.

the mcp server:
- supports streamable HTTP
- unit test coverage.

removed the legacy tool bundle.

### DIFF
--- a/.github/workflows/google-calendar-build.yaml
+++ b/.github/workflows/google-calendar-build.yaml
@@ -1,0 +1,58 @@
+name: Build release google-calendar
+
+permissions:
+  id-token: write
+  contents: read
+  packages: write
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - 'google/calendar/**'
+    branches:
+      - main
+
+jobs:
+  oss-tools-build:
+    runs-on: depot-ubuntu-22.04
+    concurrency:
+      group: google-calendar-build
+      cancel-in-progress: true
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ secrets.GHCR_USERNAME }}
+          password: ${{ secrets.GHCR_TOKEN }}
+
+      - name: Build and Push Docker image
+        uses: depot/build-push-action@v1
+        id: build-and-push
+        with:
+          project: bbqjs4tj1g
+          context: ./google/calendar
+          push: true
+          pull: true
+          platforms: linux/amd64,linux/arm64
+          tags: |
+            ghcr.io/${{ github.repository }}/google-calendar:latest
+
+      - name: Install Cosign
+        uses: sigstore/cosign-installer@v3.8.1
+        with:
+          cosign-release: 'v2.4.3'
+
+      - name: Sign Images
+        env:
+          DIGEST: ${{ steps.build-and-push.outputs.digest }}
+          TAGS: ghcr.io/${{ github.repository }}/google-calendar:latest
+        run: |
+          images=""
+          for tag in ${TAGS}; do
+            images+="${tag}@${DIGEST} "
+          done
+          cosign sign --yes ${images}
--- a/google/calendar/tests/test_server.py
+++ b/google/calendar/tests/test_server.py
@@ -8,7 +8,6 @@ import os
 # Add the parent directory to the Python path so we can import from app
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
 from app.server import mcp
-import json
 
 # Configure pytest for async support
 pytestmark = pytest.mark.asyncio


drpebcak:
Does this run cleanly with uvx? Doesn't seem like you are creating a package that we could publish or run.

tybalex:
> Does this run cleanly with uvx? Doesn't seem like you are creating a package that we could publish or run.

no, let me add it.

drpebcak:
That looks better, but i have an additional question - does obot understand streamable-http? my current understanding is that it only understand stdio for mcp servers that are obot-hosted.

drpebcak:
> @drpebcak @tybalex I'm 99% sure that Obot can do Streamable HTTP. For stdio servers, we deploy those into Kubernetes as pods, and so we must communicate with them over the network rather than pipes.

right, but obot will launch this the same way it does with a stdio server, and the supergateway config will break because this one is not stdio

g-linville:
oh okay that makes sense

tybalex:
> > @drpebcak @tybalex I'm 99% sure that Obot can do Streamable HTTP. For stdio servers, we deploy those into Kubernetes as pods, and so we must communicate with them over the network rather than pipes.
> 
> right, but obot will launch this the same way it does with a stdio server, and the supergateway config will break because this one is not stdio

hang on, I thought we only want to launch stdio servers for those 3rd party MCP servers, not Obot's `first-party` mcps. Obot's native MCP servers should be multi-tenant and supports streamable-HTTP. @cjellick can you confirm?

drpebcak:
I hadn't heard of any distinction between ours and 3rd party ones. What does 'native' mean in this context? I'd imagine that the support for these is the same as any mcp server. It would be good to work towards supporting streamable http and sse though regardless.

tybalex:
> I hadn't heard of any distinction between ours and 3rd party ones. What does 'native' mean in this context? I'd imagine that the support for these is the same as any mcp server. It would be good to work towards supporting streamable http and sse though regardless.

by `native` I mean any mcp servers converted from our legacy tool bundles.

drpebcak (changes requested):

google/calendar/pyproject.toml
```diff
@@ -11,6 +15,26 @@ dependencies = [
     "rfc3339-validator>=0.1.4",
 ]
 
+[project.scripts]
+google-calendar-mcp = "server:main"
+
+[project.urls]
+Homepage = "https://github.com/obot-platform/otto8-tools"
+Repository = "https://github.com/obot-platform/otto8-tools"
```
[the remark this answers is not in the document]
tybalex: thx, updated.

google/calendar/server.py
```diff
@@ -0,0 +1,717 @@
...
+    if (
+        not time_min and not time_max
+    ):  # if no time_min or time_max is provided, default to the current time, so it will list all upcoming events
+        time_min = get_current_time_rfc3339()
+        params["timeMin"] = time_min
+
+    if order_by:
+        params["orderBy"] = order_by
+
+    q = os.getenv("Q")
```
g-linville, line 199: Why are we reading an environment variable here?

google/calendar/server.py
```diff
@@ -0,0 +1,717 @@
...
+
+    try:
+        existing_event = (
+            service.events().get(calendarId=calendar_id, eventId=event_id).execute()
+        )
+        if (
+            existing_event_type := existing_event.get("eventType")
+        ) not in MOVABLE_EVENT_TYPES:
+            raise ValueError(
+                f"Events with type '{existing_event_type}' can not be moved."
```
g-linville, line 285:
```suggestion
                f"Events with type '{existing_event_type}' cannot be moved."
```

google/calendar/server.py
```diff
@@ -0,0 +1,717 @@
...
+
+
+@mcp.tool(
+    exclude_args=["cred_token"],
+)
+def create_event(calendar_id: Annotated[str, Field(description="Calendar id to create event in. Set to `primary` to create event in the primary calendar")],
+                 summary: Annotated[str, Field(description="Event title")] = "My Event",
+                 location: Annotated[str, Field(description="Geographic location of the event as free-form text.")] = "",
+                 description: Annotated[str, Field(description="Event description")] = "",
+                 time_zone: Annotated[str | None, Field(description="Event time zone to create. Defaults to the user's timezone. Must be a valid IANA timezone string")] = None,
```
g-linville, line 327: > `Default's to the user's timezone.`

Is this still true when this runs a as a standalone MCP server?

drpebcak: It'd be server time unless we do something with `OBOT_USER_TIMEZONE`

google/calendar/server.py
```diff
@@ -0,0 +1,717 @@
...
+# Configure server-specific settings
+PORT = os.getenv("PORT", 9000)
+MCP_PATH = os.getenv("MCP_PATH", "/mcp/google-calendar")
+
+mcp = FastMCP(
+    name="GoogleCalendarMCPServer",
+    on_duplicate_tools="error",                  # Handle duplicate registrations
+    on_duplicate_resources="warn",
+    on_duplicate_prompts="replace",
+    mask_error_details=False, # only include details for ToolError and convert other errors to generic error\
```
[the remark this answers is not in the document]
tybalex: yep, updated it to true. was  setting to False for dev

google/calendar/server.py
```diff
@@ -0,0 +1,717 @@
...
+)
+
+@mcp.tool(
+    annotations={
+        "readOnlyHint": True,
+        "destructiveHint": False,
+    },
+    exclude_args=["cred_token"],
```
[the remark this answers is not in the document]
tybalex: now removed and integrated with Daishan's Oauth proxy to get the token from custom http header

google/calendar/server.py
tybalex: thanks lol, removed. somehow this line sneaks in

google/calendar/server.py
tybalex: not really, now it defaults to the server timezone

tybalex: I hope Obot can somehow supply this information to the MCP server though

tybalex:
> Please add the github action to build this image

done.

PR #702, opened 2025-06-17
Enhance: add ability to fetch user profile

StrongMonkey:
Add a different endpoint to fetch user profile from auth provider. It can be used to supply friendly username/icon url from auth provider.


g-linville (approved):
Make sure you update this in the Okta and Entra auth providers in the enterprise tools repo as well. I don't remember if they even return an icon right now, but the paths should at least be updated so that they don't return 404s.

PR #708, opened 2025-06-24
feat: replace google drive tool bundle with Google Drive MCP server

tybalex:
replace legacy google drive tool bundle with the Google Drive MCP server.

the mcp server:
- assumes an access_token will be provided by Obot for each tool call.
- supports streamable-HTTP and stdio
- unit test coverage.
- download/upload file not supported yet. Needs workspace support.

removed the legacy tool bundle.


google/drive/apis/workspace_file.py
```diff
@@ -1,8 +1,8 @@
 import os
-import gptscript
+# import gptscript
```
thedadams, line 2: Remove if not needed.

google/drive/example_client.py
thedadams, line 1: Do we need this file? Seems like something that could be documentation.

google/drive/server.py
```diff
@@ -0,0 +1,481 @@
...
+            trashed=False,
+        )
+        
+        return files
+    except HttpError as error:
+        raise HttpError(f"Failed to list files, HttpError: {error}")
+    except Exception as error:
+        raise ToolError(f"Unexpected ToolError: {error}")
+
+## TODO: this tool requires the support of workspace
```
thedadams, line 64: Remove code that is not needed.

thedadams (dismissed):
Have we discussed how these are going to be launched? If we are implementing them as streamable HTTP servers, then we would need Docker images for them to work in Obot's Kubernetes deployments. But that wouldn't work for local stdio stuff.

tybalex:
> Have we discussed how these are going to be launched? If we are implementing them as streamable HTTP servers, then we would need Docker images for them to work in Obot's Kubernetes deployments. But that wouldn't work for local stdio stuff.

From what I've seen from Monday's  standup, the google mcp servers will be launched as remote mcp servers for Obot or Darrent's chat client. I am going to add a dockerfile so we can build Docker images.

google/drive/example_client.py
tybalex: removed

google/drive/server.py
tybalex: removed

tybalex:
> This one also needs the github action to build

done.

drpebcak (approved):

PR #709, opened 2025-06-24
chore: convert bluesky to mcp

drpebcak:


drpebcak:
Those tests are for gptscript, which has nothing to do with this.

It does not support multi-tenant, but that's not a requirement as far as I know.

PR #710, opened 2025-07-01
enhance: expose github auth group data

njhale:
- Add the `/obot-list-auth-group` endpoint to the github auth provider
  to support searching for a user's org/teams
- Enrich `/obot-get-state` response to include orgs/teams data

### DIFF
--- a/github-auth-provider/main.go
+++ b/github-auth-provider/main.go
@@ -16,6 +16,7 @@ import (
 	"github.com/obot-platform/tools/auth-providers-common/pkg/env"
 	"github.com/obot-platform/tools/auth-providers-common/pkg/state"
 	"github.com/obot-platform/tools/github-auth-provider/pkg/profile"
+	"github.com/sahilm/fuzzy"
 )
 
 type Options struct {
@@ -60,6 +61,7 @@ func main() {
 	legacyOpts := options.NewLegacyOptions()
 	legacyOpts.LegacyProvider.ProviderType = "github"
 	legacyOpts.LegacyProvider.ProviderName = "github"
+	legacyOpts.LegacyProvider.Scope = "user:email read:org"
 	legacyOpts.LegacyProvider.ClientID = opts.ClientID
 	legacyOpts.LegacyProvider.ClientSecret = opts.ClientSecret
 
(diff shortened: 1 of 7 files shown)


github-auth-provider/main.go
```diff
@@ -133,13 +135,14 @@ func main() {
...
 	mux.HandleFunc("/obot-get-user-info", func(w http.ResponseWriter, r *http.Request) {
-		userInfo, err := profile.FetchGitHubProfile(r.Context(), r.Header.Get("Authorization"), "https://api.github.com/user")
+		userInfo, err := profile.FetchUserProfile(r.Context(), r.Header.Get("Authorization"))
 		if err != nil {
 			http.Error(w, fmt.Sprintf("failed to fetch user info: %v", err), http.StatusBadRequest)
 			return
 		}
 		json.NewEncoder(w).Encode(userInfo)
 	})
+	mux.HandleFunc("/obot-list-auth-groups", listGroups(legacyOpts.LegacyProvider.GitHubToken))
```
StrongMonkey, line 145: Do we ever pass the GitHubToken into auth provider? i thought it is already coming from the obot where it fetch from state.

github-auth-provider/pkg/profile/profile.go
```diff
@@ -43,27 +49,83 @@ type githubResponse struct {
...
+}
+
+type githubTeam struct {
+	ID           int64              `json:"id"`
+	Name         string             `json:"name"`
+	Organization githubOrganization `json:"organization"`
+}
+
+func FetchUserProfile(ctx context.Context, accessToken string) (*githubUserProfile, error) {
+	result, err := makeGitHubRequest[githubUserProfile](ctx, accessToken, "user")
```
thedadams, line 65: pico-nit: not a big var of this type of generics usage because you have to explicitly state the type parameter. In my opinion, it is less readable than just doing:
```suggestion
	var result githubUserProfile
	err := makeGitHubRequest[githubUserProfile](ctx, accessToken, "user", &result)
```

Then the signature of `makeGitHubRequest` would change to `func makeGitHubRequest(ctx context.Context, accessToken, path string, result any) error` and none of the implementation of the function changes.

njhale: fair enough. I'll update the signature and usage

njhale: fixed

thedadams (approved):

PR #712, opened 2025-07-17
fix: Gmail mcp list_draft tool

tybalex:


google/gmail/obot_gmail_mcp/apis/drafts.py
```diff
@@ -13,19 +13,20 @@ async def list_drafts(service, max_results=None):
...
                 )
 
             drafts = results.get("drafts", [])
             if not drafts:
                 break
-
-            all_drafts.extend(drafts)
+            for draft in drafts:
+                draft_id, draft_str = draft_to_string(service, draft)
+                all_drafts.append(draft_str)
```
g-linville, line 29: Don't we still need to include the ID though? So that the LLM can know what it is and use it for future tool calls. If I am understanding correctly, this change would stop outputting the ID altogether.

tybalex: oh the name is kinda misleading, the `draft_str` includes the ID too.

tybalex: this function was designed like this in the past for the gptscript dataset

PR #713, opened 2025-07-17
chore: cleanup code from tools that are no longer needed

drpebcak:


g-linville (approved):
my favorite PR I've reviewed in a while

g-linville:
oh, also make sure you update index.yaml @drpebcak

njhale (changes requested):
I think we'll also want to remove the references to these from `index.yaml` as well

index.yaml
```diff
@@ -55,14 +55,10 @@ tools:
     reference: ./pagerduty
   postgres:
     reference: ./postgres
-  smartthings:
-    reference: ./smartthings
   gemini-image-generator:
```
njhale, line 58: I think this needs to be dropped too

njhale (approved):

PR #715, opened 2025-07-22
fix: switch to my fork of fastMCP to fix a custom-http-header issue.

tybalex:
this is a known fastMCP bug https://github.com/jlowin/fastmcp/issues/596, before it gets fixed, we will refer my custom fork.


tybalex:
> Is there an issue in the fastmcp repo we can reference?

yes, see the updated description.

PR #736, opened 2025-08-26
enhance: add validate tool for github auth provider

njhale:
Add a `validate` tool that can be used to validate options passed to the github-auth-provider.

The tool validates each option field and returns a JSON object containing any validation errors it encountered. Each error in the response is keyed to the associated field name with the goal of driving better form UX in Obot's UI.

Addresses part of https://github.com/obot-platform/obot/issues/4039

See https://github.com/obot-platform/obot/pull/4044 for usage and loom.

### DIFF
--- a/auth-providers-common/pkg/env/env.go
+++ b/auth-providers-common/pkg/env/env.go
@@ -1,6 +1,7 @@
 package env
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"reflect"
@@ -63,4 +64,56 @@ func LoadEnvForStruct[T any](s *T) error {
 	}
 
 	return nil
+}
+
+type ValidationError struct {
+	Err error `json:"error"`
+}
+
+func (e ValidationError) Error() string {
+	return e.Err.Error()
+}
+
+type FieldValidationError struct {
+	EnvVar    string `json:"envVar"`
+	Message   string `json:"message"`
+	Value     string `json:"value"`
+	Sensitive bool   `json:"sensitive"`
+}
+
+func (e FieldValidationError) Error() string {
+	if e.Sensitive {
+		return fmt.Sprintf("invalid environment variable %s: %s (value: %s)", e.EnvVar, e.Message, "REDACTED")
+	}
+	return fmt.Sprintf("invalid environment variable %s: %s (value: %s)", e.EnvVar, e.Message, e.Value)
+}
+
+func (e FieldValidationError) MarshalJSON() ([]byte, error) {
+	// Create a copy of the struct for JSON marshaling
+	value := e.Value
+	if e.Sensitive {
+		value = "REDACTED"
+	}
+
+	return json.Marshal(struct {
+		EnvVar    string `json:"envVar"`
+		Message   string `json:"message"`
+		Value     string `json:"value"`
+		Sensitive bool   `json:"sensitive"`
+	}{
+		EnvVar:    e.EnvVar,
+		Message:   e.Message,
+		Value:     value,
+		Sensitive: e.Sensitive,
+	})
+}
+
+type FieldValidationErrors []FieldValidationError
+
+func (e FieldValidationErrors) Error() string {
+	msgs := make([]string, len(e))
+	for i, err := range e {
+		msgs[i] = err.Error()
+	}
+	return strings.Join(msgs, "\n")
 }
--- a/github-auth-provider/main.go
+++ b/github-auth-provider/main.go
@@ -16,37 +16,29 @@ import (
 	"github.com/oauth2-proxy/oauth2-proxy/v7/pkg/validation"
 	"github.com/obot-platform/tools/auth-providers-common/pkg/env"
 	"github.com/obot-platform/tools/auth-providers-common/pkg/state"
+	"github.com/obot-platform/tools/github-auth-provider/pkg/config"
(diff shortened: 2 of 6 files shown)


auth-providers-common/pkg/env/env.go
```diff
@@ -64,3 +65,55 @@ func LoadEnvForStruct[T any](s *T) error {
...
+
+func (e ValidationError) Error() string {
+	return e.Err.Error()
+}
+
+type FieldValidationError struct {
+	EnvVar    string `json:"envVar"`
+	Message   string `json:"message"`
+	Value     string `json:"value"`
+	Sensitive bool   `json:"sensitive"`
```
njhale, line 81: Not sure passing `Value` and `Sensitive` fields back is necessary, we don't use them in the front end ATM.

github-auth-provider/pkg/config/config.go
```diff
@@ -0,0 +1,176 @@
...
+func complete(o options) (*Options, error) {
+	var (
+		validationErrors env.FieldValidationErrors
+		completedOptions = Options{
+			options: o,
+		}
+	)
+
+	if o.AuthEmailDomains != "" {
+		// TODO(njhale): Add validation for email domains
```
njhale, line 67: nit: Will remove this TODO; it's no longer valid.

github-auth-provider/pkg/config/config.go
```diff
@@ -0,0 +1,176 @@
...
+			case domain != "*" && !emailDomainRegex.MatchString(domain):
+				errorMsg = fmt.Sprintf("'%s' is not a valid email domain", domain)
+			default:
+				emailDomains[i] = domain
+				continue
+			}
+
+			// Stop after the first email domain validation error
+			break
+		}
```
njhale, line 87: Pretty sure this configuration is common to all auth providers. It has me wondering if part of this validation /completion should live in `env.LoadEnv`; e.g. you build a validator/completor to pass into `env.LoadEnv`.

github-auth-provider/pkg/config/config.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil || refreshDuration <= 0 {
+		validationErrors = append(validationErrors, env.FieldValidationError{
+			EnvVar:    "OBOT_AUTH_PROVIDER_TOKEN_REFRESH_DURATION",
+			Message:   "must be a valid duration string and greater than 0",
+			Value:     o.AuthTokenRefreshDuration,
+			Sensitive: false,
+		})
+	} else {
+		completedOptions.AuthTokenRefreshDuration = refreshDuration
+	}
```
njhale, line 112: Same comment about generic/composable validation here.

github-auth-provider/pkg/config/config_test.go
```diff
@@ -0,0 +1,533 @@
...
+		input    options
+		expected expected
+	}{
+		{
+			name: "valid configuration with all fields",
+			input: options{
+				ClientID:                 "client123",
+				ClientSecret:             "secret456",
+				ObotServerURL:            "https://example.com",
+				AuthCookieSecret:         "cookiesecret",
```
njhale, line 29: These fields aren't actually validated by the code this is testing, but I think it probably will be after the first pass goes in, so I left them in.

github-auth-provider/pkg/config/config.go
```diff
@@ -0,0 +1,176 @@
...
+			validationErrors = append(validationErrors, env.FieldValidationError{
+				EnvVar:    "OBOT_AUTH_PROVIDER_EMAIL_DOMAINS",
+				Message:   errorMsg,
+				Value:     o.AuthEmailDomains,
+				Sensitive: false,
+			})
+		} else {
+			completedOptions.AuthEmailDomains = emailDomains
+		}
+
```
thedadams, line 99: nit:
```suggestion
```

github-auth-provider/pkg/config/config.go
```diff
@@ -0,0 +1,176 @@
...
+func loadEnv() (*Options, error) {
+	var opts options
+	if err := env.LoadEnvForStruct(&opts); err != nil {
+		return nil, fmt.Errorf("failed to load environment variables: %w", err)
+	}
+
+	return complete(opts)
+}
+
+func complete(o options) (*Options, error) {
```
thedadams, line 58:
```suggestion
func completeAndValidate(o options) (*Options, error) {
```

github-auth-provider/main.go
```diff
@@ -16,37 +16,29 @@ import (
...
-	var opts Options
-	if err := env.LoadEnvForStruct(&opts); err != nil {
-		fmt.Printf("ERROR: github-auth-provider: failed to load options: %v\n", err)
-		os.Exit(1)
-	}
+	opts, err := config.LoadEnv()
+	if len(os.Args) > 1 && os.Args[1] == "validate" {
+		if err != nil {
+			var validationErr env.ValidationError
+			if errors.As(err, &validationErr) {
```
StrongMonkey, line 29: Actually, when validation fails, you should do this and return os.exit(0) https://github.com/obot-platform/enterprise-tools/pull/64/files#diff-01e49a110328ad7a46eea37ea6af13367c4de003c836988ef175f66f9750839aR27

This is so that the backend will not interpret it as an error and just read from output(json object) to decide if it is an error.

PR #742, opened 2025-09-09
Feat: Add excel mcp server

StrongMonkey:
https://github.com/obot-platform/obot/issues/4193

I did removed one tool call `query_worksheet_data` since it uses gptscript which is not compatiable any more.

This is mostly a direct copy of our existing excel tool, with converting to mcp server using go mcp sdk and leverage access token passed through mcp-oauth-proxy.


microsoft365/excel-mcp-go/main.go
```diff
@@ -0,0 +1,769 @@
...
+
+// AddWorksheetColumn adds a column to a worksheet
+func (e *ExcelMCPServer) AddWorksheetColumn(ctx context.Context, req *mcp.CallToolRequest, args AddWorksheetColumnArgs) (*mcp.CallToolResult, any, error) {
+	contents := strings.Split(args.Contents, "|")
+	re := regexp.MustCompile(`\d+$`)
+	startRow, err := strconv.Atoi(re.FindString(args.ColumnID))
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to parse starting row: %v", err)
+	}
+	endRow := startRow + len(args.Contents) - 1
```
tybalex, line 527:
```suggestion
	endRow := startRow + len(contents) - 1
```

microsoft365/excel-mcp-go/main.go
```diff
@@ -0,0 +1,769 @@
...
+		// Excel epoch starts at 1900-01-01, but there's a leap year bug
+		// Serial 1 = 1900-01-01, Serial 60 = 1900-02-28, Serial 61 = 1900-03-01
+		// This is a simplified conversion
+		epochDays := serial - 1
+		if serial > 59 {
+			epochDays-- // Account for Excel's leap year bug
+		}
+
+		// Convert to Go time and format as YYYY-MM-DD
+		date := "1900-01-01" // Simplified - would need proper date arithmetic
```
tybalex, line 633: is it always returning 1900-01-01?

tybalex (changes requested):

microsoft365/excel-mcp-go/main.go
StrongMonkey: Oh, good catch, my bad

microsoft365/excel-mcp-go/main.go
StrongMonkey: yeah, i am just converting existing method. You are right, this is not converting it correctly

PR #802, opened 2026-05-06
feat: Add Generic Responses model provider

calvinmclean:
Addresses https://github.com/obot-platform/obot/issues/6506

Rename Ollama model provider to Generic Responses and add optional API Key.

Also addresses https://github.com/obot-platform/obot/issues/6410 since errors were not working correctly for the new provider. This puts error handling more inline with the enterprise tools.

Screenshot 2026-05-06 at 16 06 04
Screenshot 2026-05-07 at 08 08 27
Screenshot 2026-05-07 at 08 11 49
Screenshot 2026-05-07 at 08 12 03

### DIFF
--- a/openai-model-provider/proxy/proxy.go
+++ b/openai-model-provider/proxy/proxy.go
@@ -59,8 +59,10 @@ func (cfg *Config) EnsureURL() error {
 		return nil
 	}
 
-	// Remove any trailing slashes from BaseURL
-	cfg.BaseURL = strings.TrimRight(cfg.BaseURL, "/")
+	cfg.BaseURL = strings.TrimRight(strings.TrimSpace(cfg.BaseURL), "/")
+	if cfg.BaseURL == "" {
+		return errors.New("base URL cannot be empty or '/'")
+	}
 
 	u, err := url.Parse(cfg.BaseURL)
 	if err != nil {
--- a/openai-model-provider/proxy/validate.go
+++ b/openai-model-provider/proxy/validate.go
@@ -2,30 +2,57 @@ package proxy
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
-	"log/slog"
+	"log"
 	"net/http"
+	"os"
+	"strings"
 
 	"github.com/obot-platform/tools/openai-model-provider/api"
 )
 
-func handleValidationError(err error, loggerPath, msg string) error {
-	slog.Error(msg, "logger", loggerPath, "error", err)
-	fmt.Printf("{\"error\": \"%s\"}\n", msg)
-	return fmt.Errorf(msg)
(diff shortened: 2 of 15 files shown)


openai-model-provider/proxy/validate.go
```diff
@@ -38,25 +65,30 @@ func (cfg *Config) Validate(toolPath string) error {
...
-			return handleValidationError(fmt.Errorf("status %s: %s", resp.Status, string(body)), toolPath, fmt.Sprintf("Invalid %s Credentials", cfg.Name))
+			if bodyBytes, readErr := io.ReadAll(io.LimitReader(resp.Body, 4096)); readErr == nil {
+				body = bodyBytes
+			}
+		}
+		if upstreamMsg := parseUpstreamErrorMessage(body); upstreamMsg != "" {
+			return handleValidationError(fmt.Errorf("status %s: %s", resp.Status, string(body)), upstreamMsg)
 		}
-		return handleValidationError(fmt.Errorf("status %s", resp.Status), toolPath, fmt.Sprintf("Invalid %s Credentials", cfg.Name))
+		return handleValidationError(fmt.Errorf("status %s", resp.Status), "")
```
thedadams: I would have guessed that in the first case, we use include the `upstreamMsg` and in the second case, we include `string(body)`.

Is there a reason to do both in the first and neither in the second?

openai-model-provider/proxy/proxy.go
```diff
@@ -59,8 +59,10 @@ func (cfg *Config) EnsureURL() error {
 		return nil
 	}
 
-	// Remove any trailing slashes from BaseURL
-	cfg.BaseURL = strings.TrimRight(cfg.BaseURL, "/")
+	cfg.BaseURL = strings.TrimRight(strings.TrimSpace(cfg.BaseURL), "/")
+	if cfg.BaseURL == "" || cfg.BaseURL == "/" {
```
thedadams: Based on the code above, it is never possible for `cfg.BaseURL == "/"`.
```suggestion
	if cfg.BaseURL == "" {
```

thedadams (approved):

openai-model-provider/proxy/validate.go
calvinmclean: Any compatible API _should_ have the upstreamMsg.
In the first case, we are logging body and including the upstreamMsg in the response that goes to the UI.
In the second case, I don't want to put the whole body as the message since it can be large and break the UI's regex for extracting an error.

However, I should still be logging the body here. I will double-check some things and then fix/improve

calvinmclean: Ok I improved this a bit and added some tests

g-linville (approved):

